PR #155, opened 2026-06-28
2. [Fix] Reapply persisted provider override during service boot

unohee:
## Summary
Update service bootstrap after web.setWebRunner to readProviderOverride() and reapply it via setDefaultAdapter/runnerInstance.switchProvider when it differs from config.adapter default. Keep boot no-op when override matches current adapter. Add/adjust service test coverage if existing harness supports it. Completion criteria: daemon startup restores codex override even if config.yaml was reset.

* Depends on: 1. [Fix] Restore provider override persistence module
* File scope: src/core/service.ts, src/core/service.test.ts
* Estimate: 35 min

*Split out from "fix(provider): codex 고정이 새는 두 회귀 — provider-override 부트 미적용 + jobProfile model 미전달 → Codex-Spark 자체 한도 소진" during planning.*

## Linear
Closes INT-1981

---

### DIFF
--- a/src/core/service.test.ts
+++ b/src/core/service.test.ts
@@ -10,6 +10,10 @@ import {
 } from './service.js';
 import { setDefaultAdapter } from '../adapters/index.js';
 
+const mockRunnerInstance = {
+  switchProvider: vi.fn(),
+};
+
 // Mock external dependencies
 // Mock auth so service.test never reads the real ~/.openswarm/auth-profiles.json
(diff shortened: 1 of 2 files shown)


unohee:
Superseded by PR #153 (INT-1979) which already implements service.ts boot re-apply. Duplicate from daemon decomposition of INT-1979 — closing per INT-1985 cleanup.

PR #182, opened 2026-07-01
Constrain project cancellation path matching

ag-linden:
## Summary

This tightens `TaskScheduler.cancelProjectTasks()` so disabling a project only aborts tasks whose running path is the exact project path or a real descendant path.

Previously the matcher used raw string prefixes. That preserved worktree cancellation, but it also meant disabling `/dev/WAVE` could abort an unrelated running task under a sibling path like `/dev/WAVE-next`.

Changes:

- normalize scheduler cancellation paths before comparison
- match exact project paths or child path boundaries only
- add a regression test covering the sibling-prefix case

## Related issue

No linked issue; this is a small scheduler lifecycle bug fix.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Docs

## Validation

- [x] `npm test -- src/orchestration/taskScheduler.cancel.test.ts`
- [x] `npm test -- src/orchestration/taskScheduler.cancel.test.ts src/orchestration/taskScheduler.concurrency.test.ts`
- [x] `npm run lint` (passes with existing warnings outside this change)
- [x] `npm run typecheck`
- [x] `npm run build`
- [ ] `npm test` passes (added/updated tests where relevant)

Full `npm test` currently has unrelated failures that reproduce standalone on this checkout:

- `src/automation/autonomousRunner.cancel.test.ts`: hook timeout while importing runner modules
- `src/automation/autonomousRunner.infraError.test.ts`: hook timeout while importing runner modules
- `src/core/service.test.ts`: provider override expectations fail (`setDefaultAdapter`/`readProviderOverride`)
- `src/tui/components/AuditBoard.test.tsx`: errored area tally expectation does not update to `1 failed`

The scheduler cancellation tests covering this patch pass.

### DIFF
--- a/src/orchestration/taskScheduler.cancel.test.ts
+++ b/src/orchestration/taskScheduler.cancel.test.ts
@@ -47,6 +47,17 @@ describe('TaskScheduler cancellation', () => {
     expect(c.wasAborted()).toBe(false);
   });
 
+  it('cancelProjectTasks does not abort sibling paths with the same prefix', () => {
+    const a = deferredExecutor();
+    const b = deferredExecutor();
+    sched.startTask(task('a'), '/dev/WAVE', a.exec);
+    sched.startTask(task('b'), '/dev/WAVE-next', b.exec);
+    const n = sched.cancelProjectTasks('/dev/WAVE');
+    expect(n).toBe(1);
+    expect(a.wasAborted()).toBe(true);
+    expect(b.wasAborted()).toBe(false);
+  });
+
   it("a cancelled result is not counted as failed", async () => {
     const d = deferredExecutor();
     const events: string[] = [];
--- a/src/orchestration/taskScheduler.ts
+++ b/src/orchestration/taskScheduler.ts
@@ -7,6 +7,17 @@ import { EventEmitter } from 'node:events';
 import type { TaskItem } from './decisionEngine.js';
 import type { PipelineResult } from '../agents/pairPipeline.js';
 
+function normalizeProjectPath(path: string): string {
+  const normalized = path.replace(/\\/g, '/').replace(/\/+$/g, '');
+  return normalized || '/';
+}
+
+function isSameProjectOrDescendant(candidatePath: string, projectPath: string): boolean {
+  const candidate = normalizeProjectPath(candidatePath);
+  const project = normalizeProjectPath(projectPath);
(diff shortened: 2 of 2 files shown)


unohee:
Post-merge review — LGTM ✅

Thanks for the clean fix and the regression test, @ag-linden. I re-reviewed the merged code on `origin/main` and it holds up well.

**Correctness**
- The core bug — `p.startsWith(projectPath)` treating `/dev/WAVE-next` as part of `/dev/WAVE` — is correctly fixed by the `${project}/` boundary in `isSameProjectOrDescendant`, and your new test pins exactly that case.
- Worktree cancellation is preserved: the existing `cancelProjectTasks aborts every task on the project (repo + worktree paths)` test (which aborts `/dev/WAVE/worktree/abc` via `/dev/WAVE`) still passes against the new matcher, so the descendant branch does its job.
- Dropping the old third condition (`projectPath.startsWith(p)`, the *ancestor* direction) is the right call — that branch would abort a **parent** project's in-flight task when disabling a child path, which was a latent over-cancellation bug rather than intended behavior. In practice `cancelProjectTasks` is always called with the base project path (`disableProject` passes the same path that was `enqueue`d), so the exact-match arm carries the real case and the descendant arm covers worktree subpaths.
- `normalizeProjectPath` (backslash→`/`, trailing-slash strip) applied symmetrically to both operands is a nice touch.

Verified locally on the merged code: `taskScheduler.cancel.test.ts` (7) and `taskScheduler.concurrency.test.ts` (3) both green.

**Optional follow-ups (non-blocking):**
- `normalizeProjectPath` doesn't expand `~` or resolve relative paths. That's fine today because the enqueue side and `disableProject` both receive the same config-sourced path representation, so exact-match lines up — but if those two ever diverge (one `~/dev/WAVE`, one `/Users/.../dev/WAVE`) cancellation would silently no-op. A `resolve()` / `~`-expansion in the normalizer would be a reasonable belt-and-suspenders later.
- On the unrelated full-suite failures you flagged: `src/core/service.test.ts` is indeed pre-existing (fails on `main` too, unrelated to this change). The two `autonomousRunner` import-timeout tests and `AuditBoard.test.tsx` passed in my run, so those look like environment/timing (module-import hook timeout) rather than anything your patch touched — consistent with your "reproduce standalone" note.

Thanks again for the contribution! 🙏

PR #192, opened 2026-07-01
Normalize project cancellation paths

ag-linden:
## Summary

Follow-up to the optional note on #182.

This makes project cancellation path matching tolerant of equivalent path representations before applying the exact-or-descendant boundary check.

Changes:

- expand `~` in scheduler cancellation paths
- resolve relative paths before comparison
- keep the existing slash normalization and trailing-slash trimming
- add regression coverage for home-expanded and relative disable paths

## Related issue

Follow-up to #182.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Docs

## Validation

- [x] `npm test -- src/orchestration/taskScheduler.cancel.test.ts`
- [x] `npm test -- src/orchestration/taskScheduler.cancel.test.ts src/orchestration/taskScheduler.concurrency.test.ts`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run build`
- [ ] `npm test` passes

Local full `npm test` still has unrelated environment/timing failures outside this scheduler change. The autonomous runner files passed when reproduced standalone; `src/tui/components/AuditBoard.test.tsx` still fails locally on the errored-area tally expectation.

### DIFF
--- a/src/orchestration/taskScheduler.cancel.test.ts
+++ b/src/orchestration/taskScheduler.cancel.test.ts
@@ -1,4 +1,6 @@
 import { describe, it, expect, beforeEach } from 'vitest';
+import { homedir } from 'node:os';
+import { resolve } from 'node:path';
 import { TaskScheduler } from './taskScheduler.js';
 import type { TaskItem } from './decisionEngine.js';
 import type { PipelineResult } from '../agents/pairPipeline.js';
@@ -56,6 +58,22 @@ describe('TaskScheduler cancellation', () => {
     expect(n).toBe(1);
     expect(a.wasAborted()).toBe(true);
     expect(b.wasAborted()).toBe(false);
+  });
+
+  it('cancelProjectTasks matches home-expanded project paths', () => {
+    const d = deferredExecutor();
+    sched.startTask(task('home'), resolve(homedir(), 'dev/WAVE'), d.exec);
+    const n = sched.cancelProjectTasks('~/dev/WAVE');
+    expect(n).toBe(1);
+    expect(d.wasAborted()).toBe(true);
+  });
+
+  it('cancelProjectTasks matches relative project paths', () => {
+    const d = deferredExecutor();
+    sched.startTask(task('relative'), resolve(process.cwd(), 'relative-WAVE'), d.exec);
+    const n = sched.cancelProjectTasks('./relative-WAVE');
+    expect(n).toBe(1);
+    expect(d.wasAborted()).toBe(true);
   });
 
   it("a cancelled result is not counted as failed", async () => {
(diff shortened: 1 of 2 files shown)


unohee (approved):
Reviewed the diff and ran it locally on top of current main: the 2 new regression tests + all 40 orchestration tests pass, tsc clean. CI is green.

The change is correct and nicely scoped:

- `~`/`~/` expansion and `resolve()` are applied inside `normalizeProjectPath`, so **both** sides of the comparison (stored task path and cancellation query) go through the same canonicalization — no asymmetry.
- Re-normalizing backslashes *after* `resolve()` keeps the Windows behavior of the original slash normalization. Good catch.
- Bonus win worth calling out: `resolve()` collapses `.`/`..` segments, which actually **fixes a latent descendant-check bypass** — before this, disabling `/dev/WAVE` would also cancel a task registered under `/dev/WAVE/../WAVE-next` (it string-prefix-matched `/dev/WAVE/`). Verified: it now normalizes to `/dev/WAVE-next` and is correctly left alone.

One non-blocking edge to be aware of: `normalizeProjectPath('')` used to normalize to `'/'` (matched nothing in practice), but `resolve('')` returns `process.cwd()` — so an accidental empty-string cancellation would now cancel tasks under the daemon's cwd. Today's only production caller (`disableProject`) always passes a concrete repo path, so this is theoretical, but a one-line guard would make it airtight:

```ts
function normalizeProjectPath(path: string): string {
  if (!path.trim()) return '/';
  ...
}
```

Happy to take that as a follow-up on our side if you'd rather keep this PR as-is. Thanks for the careful follow-through on the #182 note — including the regression tests for both forms.

(The local `AuditBoard.test.tsx` failure you mentioned is unrelated to this change — it passes in CI and on current main.)

unohee:
Merged — thanks @ag-linden! This is exactly the kind of follow-through we love: you took an optional review note from #182 and came back with a tight, well-tested fix. A few things that stood out while reviewing:

- Putting the expansion/resolution inside `normalizeProjectPath` (so both comparison sides canonicalize identically) was the right call — no asymmetry to reason about.
- Re-applying the slash normalization *after* `resolve()` preserved the Windows behavior. Easy to miss.
- Unplanned bonus: `resolve()` collapsing `.`/`..` segments fixed a latent descendant-check bypass (`/dev/WAVE/../WAVE-next` used to be cancelled by disabling `/dev/WAVE`). We verified this on top of current main.

The one theoretical edge from the review (`resolve('')` → cwd) is handled in follow-up #197, so nothing needed on your side. Both changes will ship in the next release with credit in the changelog. Looking forward to your next PR 🙌

PR #275, opened 2026-07-10
test(ci): enforce coverage ratchet (INT-2671)

unohee:
## Summary
- raise measured unit coverage from 46.6% baseline to 92.95% lines
- enforce 90/90/90/80 coverage thresholds in CI
- harden CI/release/review workflows with timeouts, concurrency, and duplicate-publish checks

## Validation
- `npm test -- --run`: 2636 passed, 1 skipped
- `npm run test:coverage`: lines 92.95%, statements 91.94%, functions 92.34%, branches 83.75%
- `npm run typecheck`: passed
- `npm run lint`: 0 warnings, 0 errors
- `npm run build`: passed
- `git diff --check`: passed

## Review gate
- `openswarm review --path .` was invoked but hung after beginning analysis for over 90 seconds; interrupted with exit 130. The workflow timeout added here addresses this shared-runner failure mode.

Linear: INT-2671

PR #277, opened 2026-07-10
fix(worktree): prune stale retry metadata (INT-2669)

unohee:
## Summary
- prune stale Git worktree admin metadata after retry fallback removal
- make path containment reject only real parent segments, not safe names such as `...`
- update real-Git regression tests for both repaired behaviors

## Validation
- `npx vitest run src/support/worktreeManager.coverage.test.ts src/support/worktreeManager.test.ts`: 61 passed
- `npm run typecheck`: passed
- `git diff --check`: passed

## Stack
Based on #275 because the regression coverage file is introduced there. Merge #275 first.

Linear: INT-2669


unohee:

PR #278, opened 2026-07-10
fix(fanout): ignore injected shared symlinks (INT-2670)

unohee:
## Summary
- track shared-path symlinks injected into each fanout sandbox
- exclude only those harness-created paths from candidate scoring and promotion
- keep legitimate candidate edits promotable while preventing `EISDIR` on directory-only gitignore patterns

## Validation
- `npx vitest run src/agents/workerFanout.coverage.test.ts src/agents/workerFanout.test.ts`: 15 passed
- `npm run typecheck`: passed
- `git diff --check`: passed

## Stack
Based on #275 because the regression coverage file is introduced there. Merge #275 first.

Linear: INT-2670


unohee:

PR #280, opened 2026-07-10
feat(verify): discover repository commands (INT-2660)

unohee:
## Summary
- discover deterministic verify commands from Node/TypeScript, Python, Rust, and Go repository signals
- ignore npm placeholder tests and tolerate malformed package metadata
- keep discovery filesystem-only; no command execution

## Validation
- `npx vitest run src/verify`: 18 passed
- includes an OpenSwarm checkout smoke case
- `npm run typecheck`: passed
- `npm run lint`: 0 warnings, 0 errors
- `git diff --check`: passed

## Stack
Based on #279 for the exported `VerifyCommand` type. Merge #279 first.

Linear: INT-2660

PR #279, opened 2026-07-10
feat(verify): add repository manifest loader (INT-2658)

unohee:
## Summary
- add strict Zod schemas for repository verify manifests and commands
- load `.openswarm/verify.yaml` with explicit absence, parse, read, and validation outcomes
- provide an English commented example manifest

## Validation
- `npx vitest run src/verify`: 9 passed
- `npm run typecheck`: passed
- full `npm test -- --run`: 1805 passed, 1 skipped
- `npm run lint`: 0 warnings, 0 errors
- `git diff --check`: passed

Linear: INT-2658

PR #281, opened 2026-07-10
feat(verify): add baseline diff runner (INT-2661)

unohee:
## Summary
- run each verify command against head and only run a detached merge-base worktree when head fails
- mark only head-fail/base-pass as a new blocking failure
- classify spawn, command-not-found, and wall-clock timeout failures as infrastructure
- cap combined command output to the last 8KB and always clean temporary worktrees

## Validation
- `npx vitest run src/verify`: 24 passed
- runner fixture repeated 3x; all passed
- full `npm test -- --run`: 1821 passed, 1 skipped
- `npm run typecheck`: passed
- `git diff --check`: passed

## Stack
Based on #280 for manifest/discovery types. Merge #279 then #280 first.

Linear: INT-2661

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -4,10 +4,13 @@ name: OpenSwarm Review
 # the daemon's own swarm/* branches) and posts the verdict as a PR comment under
 # the openswarm-review[bot] identity. (INT-2552)
 #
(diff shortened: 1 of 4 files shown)


unohee:

PR #331, opened 2026-07-26
fix(discord): deliver long output and match replies to the message they answer

unohee:
## Summary

Three Discord delivery defects from the audit backlog. All are invisible in isolation and only bite with real traffic.

### 1. Long output never arrived at all

`sendToChannel` split at **3900** characters. Discord's hard limit on message *content* is **2000** — 4096 is the embed *description* limit, and a comment conflated the two. The consequence is worse than a truncated message:

| Content length | What happened |
|---|---|
| 2001–3900 | not split, rejected by the API |
| > 3900 | split into 3900-char chunks, **every chunk rejected** |

So the splitter never produced a valid chunk. `sendToThread` in the same file already used 1900 correctly; both now share one constant, so the two cannot drift apart again.

### 2. Replies attached to the wrong message

`updateLastHistoryResponse` wrote to the **last** entry in the channel's history. Two messages in one channel are handled concurrently and the model does not finish them in arrival order, so whichever finished first overwrote the newer message's slot — the newer message ended up with no answer of its own, and the older one displayed a reply to a question nobody asked.

Now matched by `messageId`, which `HistoryEntry` already carried. A response whose entry has aged out of the ring buffer is **dropped rather than misfiled**, since attaching it to an unrelated message is worse than losing it.

### 3. Stale progress after a failure

`handleDev` arms a 10s progress timer from a callback and awaited `runDevTask` with no `try/catch`. A rejection propagated out with the timer still scheduled, so ten seconds after the user was told the task failed, an "in progress" reply arrived quoting output from the run that had already ended. Cleanup moved into a `finally`, with a `settled` flag covering the window where the timer callback is already running.

## Verification

| Mutation | Test that went red |
|---|---|
| chunk size back to 3900 | 6 cases across `splitForDiscord` |
| match last entry instead of `messageId` | `attaches each response to the message it answers`, `drops a response whose entry has aged out` |
| remove the `finally` cleanup and `settled` guard | `does not post progress after the task rejects` |

The timer suite includes a **still-running** case, so the fix cannot pass by disabling progress reporting altogether — which is the cheap way to make the other two tests green.

`splitForDiscordForTest` exports the splitter bound to the production constant rather than letting tests restate the number. Restating it is exactly what allowed 3900 to sit unnoticed next to a 2000-char limit.

- Full suite: **249 files, 3321 passed, 1 skipped**
- `openswarm review`: **APPROVE**, 0 issues

### DIFF
--- a/src/discord/discordCore.ts
+++ b/src/discord/discordCore.ts
@@ -90,13 +90,35 @@ export function appendHistoryEntry(channelId: string, entry: HistoryEntry): void
 }
 
 /**
- * Add response to the last history entry
+ * Attach a response to the history entry it answers.
+ *
+ * Matched by messageId, not by position. Two messages in one channel are
+ * handled concurrently, and the model does not finish them in arrival order —
+ * writing to the last entry meant whichever finished first overwrote the newer
+ * message's entry, so the newer message's own reply was lost and the older one
+ * showed an answer to a question nobody asked.
+ *
+ * The positional fallback covers entries recorded without a messageId, which is
+ * the pre-existing behaviour for those and is safe when nothing else is in
+ * flight.
  */
-export function updateLastHistoryResponse(channelId: string, response: string): void {
+export function updateHistoryResponse(channelId: string, messageId: string | undefined, response: string): void {
   const history = channelHistoryMap.get(channelId);
-  if (history && history.length > 0) {
-    history[history.length - 1].response = response;
+  if (!history || history.length === 0) return;
+
+  if (messageId) {
+    for (let i = history.length - 1; i >= 0; i--) {
+      if (history[i].messageId === messageId) {
+        history[i].response = response;
+        return;
+      }
+    }
+    // The entry aged out of the ring buffer while the model was running.
+    // Writing to whatever is last now would attach it to an unrelated message.
+    return;
   }
+
+  history[history.length - 1].response = response;
 }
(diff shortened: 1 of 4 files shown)


unohee:
Confirmed against `src/support/dev.ts:213` — `runDevTask` registers the child's stdout/close listeners and returns `{taskId, path}` immediately, so the `finally` set `settled` before the first chunk arrived and suppressed progress for the whole run. The fix silenced the feature it was meant to protect.

Corrected in 2d1e558. The disarm now lives in `onComplete`, which `dev.ts` fires for both the `close` and `error` process events. The two paths that never reach `onComplete` are handled explicitly in `handleDev`'s own body: a synchronous `{error}` return (time window, unknown repo, task already running — no child exists) and a throw before the child is registered.

You were also right about why the test missed it: the mock made `runDevTask` block until the test released it, which describes a function this codebase does not have. Every mock now follows the real contract — return immediately, then drive `onProgress`/`onComplete` — and `posts progress for a task that is still running after handleDev returns` plus `keeps posting progress across several chunks` both fail against the previous commit, verified by mutation.

Added one case your comment surfaced indirectly: a chunk arriving *after* completion arms a brand new timer with nothing left to cancel it, so the `settled` flag is load-bearing separately from `clearTimeout`. That case fails when the flag alone is removed.

PR #333, opened 2026-07-26
fix(agents): three defects that returned the wrong answer without failing

unohee:
## Summary

Three defects from the audit backlog that share a shape: **none of them threw, logged, or failed a check.** Each just returned the wrong answer, which is why they survived several audit rounds.

### 1. A blocking guard rejected correct work

`literalExistsInHeadSource` asks git whether a contract literal exists in HEAD via `git grep -F -l <literal>`. Without `-e`, git parses a literal beginning with `-` as an **option** and exits with `unknown option`. That lands in the catch, which reports "not present in HEAD" — and this guard is `blocking`, so a contract literal like `--x-trace-id:` made the pipeline reject correct work.

Measured directly before fixing:

```
$ git grep -F -l "--output-format" HEAD -- .
error: unknown option `output-format'

$ git grep -F -l -e "--output-format" HEAD -- .
HEAD:src/adapters/base.ts   …
```

The **other** `git grep` call in the same file already passed `-e`. The correct form was sitting next to the broken one — the same shape as the Discord 3900-vs-1900 split and the `repos.json` atomic write earlier in this backlog.

### 2. Cancellation was never recorded

`cancelSession` wrote `"Session has been cancelled."` *after* transitioning to a terminal status. That transition archives the session, archiving deletes it from the live map, and `addMessage`'s lookup then misses and returns silently — so the notice never reached the transcript the user reads. The two calls are now reversed; `archiveSession` stores the same object reference, so a message added first is carried into the history.

### 3. Bus messages could be read half-written

`agentBus.publish` wrote message files in place while three consumers poll the directory with `readdir` + `readFile`. An in-place write makes the filename visible before the contents are complete. Now uses the shared atomic helper. (The temp file's `.tmp` suffix is excluded by the existing `.endsWith('.json')` filters — checked all three consumers.)

## Two tests needed a second attempt

**The atomicity test was vacuous.** Polling in a loop could not catch the window — plain `writeFile` passed **3 runs out of 3**. It now watches the directory and reads each file the instant its name appears, which is precisely what an in-place write exposes and a rename does not. Against plain `writeFile` it fails **3 out of 3**.

**The git guard test proved the wrong thing.** The first version ran `git grep` directly, so it demonstrated git's behaviour rather than the code's — removing `-e` from the source left it **green**. It now drives `runGuards` against a real repository fixture, and includes the inverse case (a literal that appears only in the test *must* still be flagged) so "stop blocking correct work" cannot be satisfied by no longer checking.

## Verification

| Mutation | Test that went red |
|---|---|
| drop `-e` from `git grep` | `does not flag a literal that HEAD already declares` |
| restore the `cancelSession` ordering | `records why the session ended` |
| `atomicWriteFile` → `fs.writeFile` | `never exposes a message file before its contents are complete` (3/3 runs) |

- Full suite: **250 files, 3333 passed, 1 skipped**
- `openswarm review`: **APPROVE**, 0 issues

---

> [!NOTE]
> **Low Risk**
> Targeted bug fixes with strong regression tests; behavior changes are narrowly scoped to guards, session cancel, and bus persistence.
> 
> **Overview**
> Fixes three agent-pipeline bugs that **failed silently** (wrong outcome, no throw or log) and adds regression tests in `silentWrongAnswers.test.ts`.
> 
> **Contract evidence guard:** `literalExistsInHeadSource` now passes `-e` to `git grep` so literals starting with `-` (e.g. `--x-trace-id:`) are not parsed as options; previously git errored, the catch reported “not in HEAD,” and the **blocking** guard rejected valid work.
> 
> **Session cancel:** `cancelSession` appends the cancellation system message **before** `updateSessionStatus('cancelled')`, so the notice is not dropped when archiving removes the session from the live map.
> 
> **Agent bus:** `AgentBus.publish` writes message JSON via `atomicWriteFile` instead of in-place `writeFile`, so directory pollers never see a `.json` name with truncated content; owner-only permissions are preserved.
> 
> Tests drive `runGuards` on real git fixtures, assert archived cancel transcripts, and race-publish large payloads against `fs.watch` + directory rescans.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9407aa1448886c0cb6f51f6864c7b3109e913fc4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agents/agentBus.ts
+++ b/src/agents/agentBus.ts
@@ -7,6 +7,7 @@ import { resolve } from 'path';
 import { homedir } from 'os';
 import * as fs from 'fs/promises';
 import { existsSync } from 'fs';
+import { atomicWriteFile } from '../support/atomicFile.js';
 
 // Types
 
@@ -168,9 +169,13 @@ export class AgentBus {
       payload,
     };
 
-    // Save message
+    // Save message. Atomic (write-temp + fsync + rename) because the consumers
+    // here poll with readdir + readFile: an in-place write makes the filename
+    // visible to readdir before its contents are complete, so a poller can read
+    // a truncated message and fail to parse it. rename publishes the name and
+    // the content in one step.
     const messagePath = resolve(this.messagesPath, `${message.id}.json`);
-    await fs.writeFile(messagePath, JSON.stringify(message, null, 2));
+    await atomicWriteFile(messagePath, JSON.stringify(message, null, 2), 0o644);
     if (++this.publishedSincePrune >= 100) {
       this.publishedSincePrune = 0;
       await this.pruneMessages();
--- a/src/agents/agentBus.ts
+++ b/src/agents/agentBus.ts
@@ -174,8 +174,14 @@ export class AgentBus {
     // visible to readdir before its contents are complete, so a poller can read
     // a truncated message and fail to parse it. rename publishes the name and
     // the content in one step.
+    //
+    // Owner-only (the helper's default), matching context.json beside it. The
+    // mode is not incidental here: atomicWriteFile chmods explicitly after the
+    // rename, so passing a wider mode would override even a restrictive umask
+    // that the previous plain writeFile respected — and these payloads carry
+    // agent prompts, outputs and errors.
     const messagePath = resolve(this.messagesPath, `${message.id}.json`);
-    await atomicWriteFile(messagePath, JSON.stringify(message, null, 2), 0o644);
+    await atomicWriteFile(messagePath, JSON.stringify(message, null, 2));
     if (++this.publishedSincePrune >= 100) {
       this.publishedSincePrune = 0;
       await this.pruneMessages();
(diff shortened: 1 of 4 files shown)


unohee:
Both correct — fixed in 9407aa1.

**Message file mode.** This was a regression I introduced, and your reasoning about `chmod`-after-rename is exactly why it mattered: the previous plain `writeFile` respected the process umask, so `0o644` did not merely keep the old behaviour, it widened it past a restrictive umask. Now on the helper's owner-only default, matching `context.json` beside it. The `0o644` came from copying the decision made for `openswarm.json`, which is a tracked repo file meant to be readable — a different situation that I carried over without re-deriving.

**`fs.watch` filename.** Also right, and it would have failed for a reason unrelated to what the test checks. The watcher is now only a trigger: every event rescans the directory, a 1ms interval backs it up so delivery is not required at all, and the assertion waits until all four files are observed intact rather than sleeping a fixed 300ms.

Both re-measured by mutation: plain `writeFile` fails 3 runs out of 3, and widening the mode back to `0o644` fails the new `keeps message files owner-only` test.

PR #337, opened 2026-07-26
fix(adapters): stop the prompt file and Add File op from clobbering and leaking

unohee:
## Summary

Two write paths that destroyed or exposed data without reporting anything.

### 1. Every CLI prompt went to a predictable path in shared `/tmp`

`spawnCli` wrote to `/tmp/openswarm-prompt-${Date.now()}.txt`. Three problems in one line:

| | Consequence |
|---|---|
| Millisecond resolution | Workers run in parallel; two calls in the same millisecond overwrote each other's prompt — and that path is handed to the CLI, so **one agent executed the other's task** |
| Default file mode | Task content readable by every local user |
| Predictable name, world-writable directory | Another local user can pre-create it as a symlink before the write lands |

`mkdtemp` answers all three: a unique **0700** directory created atomically by the OS, a **0600** file inside it, and the whole directory removed on the way out — including when the run fails.

### 2. "Add File" silently replaced existing files

The op wrote unconditionally. Naming an existing file destroyed its contents with **no error raised and nothing recorded as an update** — the agent believed it had created a file when it had replaced one. It now opens with `wx` and reports the path, which runs the existing rollback.

## Verification

| Mutation | Tests that went red |
|---|---|
| prompt back to `/tmp/…${Date.now()}.txt` | 4 (isolation, permissions, cleanup, cleanup-on-failure) |
| add op back to an unconditional write | 3 (refuses, explains, rolls back) |

One test detail worth noting: **permissions are recorded from inside `buildCommand`, not after `spawnCli` returns.** Cleanup removes the directory, so checking afterwards would have been asserting on a path that no longer exists — the first version of that test failed for exactly that reason and was corrected.

- Full suite: **253 files, 3365 passed, 1 skipped**
- `openswarm review`: **APPROVE**, 0 issues (also verified `applyV4APatch`'s sole caller surfaces the new error to the model)

---

> [!NOTE]
> **Medium Risk**
> Touches agent execution (prompt paths) and filesystem mutation/rollback in `applyV4APatch`; behavior changes are intentional safety fixes but affect all CLI spawns and codex patch application.
> 
> **Overview**
> **CLI prompts** no longer use a shared, predictable `/tmp/openswarm-prompt-${Date.now()}.txt` path. `spawnCli` now writes each prompt into a unique `mkdtemp` directory with a **0600** file and removes the whole directory in `finally`, including on failure—fixing parallel workers clobbering each other’s tasks, world-readable prompts, and symlink races.
> 
> **V4A patch apply** is hardened so failed applies cannot delete or overwrite unrelated data. Snapshots distinguish **absent**, **readable content**, and **opaque** (exists but unreadable); rollback only removes paths this patch actually created, and skips opaque paths. **Add File** uses `wx` and errors if the path exists (with guidance to use update). **Move to** refuses when the destination already exists. New tests in `clobberingWrites.test.ts` cover isolation, permissions, rollback, and refusal paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83da73929c6e65a88012716ca778000ac5b8c9a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/adapters/applyPatch.ts
+++ b/src/adapters/applyPatch.ts
@@ -119,18 +119,38 @@ export async function applyV4APatch(
     touched.add(resolvePath(op.filePath));
     if (op.moveTo) touched.add(resolvePath(op.moveTo));
   }
-  const snapshots = new Map<string, string | null>(); // abs -> prior content, or null if absent
+  // Existence and readability are recorded separately. Collapsing them into
+  // "content, or null" made rollback treat a file it could not read as one that
+  // had never existed, and rollback deletes those to restore absence — so a
+  // mode-000 file, a dangling symlink or a path owned by another user could be
+  // removed by a patch that only ever refused to touch it.
+  type Snapshot =
+    | { kind: 'absent' }
+    | { kind: 'content'; text: string }
+    | { kind: 'opaque' }; // exists, but its contents could not be captured
+  const snapshots = new Map<string, Snapshot>();
   for (const abs of touched) {
-    snapshots.set(abs, await fs.readFile(abs, 'utf-8').catch(() => null));
+    const exists = await fs.lstat(abs).then(() => true, () => false);
+    if (!exists) {
+      snapshots.set(abs, { kind: 'absent' });
+      continue;
+    }
+    const text = await fs.readFile(abs, 'utf-8').catch(() => null);
+    snapshots.set(abs, text === null ? { kind: 'opaque' } : { kind: 'content', text });
   }
   const rollback = async () => {
-    for (const [abs, content] of snapshots) {
+    for (const [abs, snapshot] of snapshots) {
       try {
-        if (content === null) {
+        if (snapshot.kind === 'opaque') {
+          // Nothing can be restored and nothing should be destroyed: leaving it
+          // as-is is the only option that cannot make the tree worse.
+          continue;
+        }
+        if (snapshot.kind === 'absent') {
           await fs.rm(abs).catch(() => {});
         } else {
           await fs.mkdir(path.dirname(abs), { recursive: true });
-          await fs.writeFile(abs, content, 'utf-8');
+          await fs.writeFile(abs, snapshot.text, 'utf-8');
         }
       } catch { /* best-effort restore */ }
     }
--- a/src/adapters/applyPatch.ts
+++ b/src/adapters/applyPatch.ts
@@ -146,7 +146,21 @@ export async function applyV4APatch(
       }
       if (op.kind === 'add') {
         await fs.mkdir(path.dirname(abs), { recursive: true });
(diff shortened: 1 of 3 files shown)


unohee:
Both fixed in dcd42f1 — and the P1 was worse than "a latent edge case", so thank you for catching it.

**Rollback deleting unreadable paths.** Verified by reverting the snapshot to its previous single-`null` form: the new `does not delete a file it could not read while refusing to add it` and `does not delete a dangling symlink while refusing to add over it` cases both go red, i.e. the file really was being removed. The guard I added in the previous commit turned a *prevented overwrite* into a *destroyed file*, which is strictly worse than the problem it was solving.

Existence is now established with `lstat` and readability recorded separately. An `opaque` entry — exists, contents not capturable — is left untouched on rollback: nothing can be restored and nothing should be destroyed, so leaving it alone is the only outcome that cannot make the tree worse.

**Prompt write inside the cleanup scope.** Also correct — a write that fails partway exited before the `try`/`finally` and left the private directory, and possibly a fragment of the prompt, behind. Moved inside.

Full suite 253 files / 3367 passed, `openswarm review` approve.

unohee:
Fixed in 9329634 — correct, and the window is real.

Removal is now limited to a `created` set: an add whose `wx` open succeeded, a move destination, and an add that failed for a reason *other* than EEXIST (which may have left a partial file this patch did create). An add refused with EEXIST is deliberately not recorded — the file is someone else's however it got there.

Worth noting because it was my first attempt and it was wrong: marking every path the loop *reached* still deleted the raced-in file, because the failing op had reached it. The distinction that matters is not "did we try this path" but "did we write it".

The race is encoded deterministically rather than by timing — the test's `resolvePath` creates the file on its second call for a path, which is exactly the window between the snapshot pass and the apply loop. Removing the `created` guard turns that test red.

All three protections in this file are now verified independently by mutation: dropping the created guard, marking EEXIST as created, and collapsing the snapshot back to a single `null` each turn a different case red. Full suite 253 files / 3368 passed.

unohee:
Fixed in 83da739 — and you're right that it was worse than the add case, for exactly the reason given: a destination that cannot be snapshotted is skipped by rollback, so the overwrite outlives a failure that reports nothing was applied.

Move now follows the same rule as Add File: if the destination exists, refuse and name the path. Self-moves are exempt and ordinary renames to a free path are unaffected — the existing move tests pass unchanged.

One of the two new tests started out **non-discriminating** and was corrected before commit. It used a mode-000 destination, but with no write permission the OS refuses the write anyway, so it passed with or without the guard. It now uses `0o200` — writable but not readable — which is the state that actually separates the two behaviours. Both tests now go red when the guard is removed.

Full suite 253 files / 3370 passed, `openswarm review` approve.

PR #345, opened 2026-07-26
fix: stop three transient failures from lasting the whole process

unohee:
## Summary

Three places where a momentary problem was preserved for the life of the process.

| | Transient cause | Permanent consequence |
|---|---|---|
| **Provider request** | connection accepts, then goes silent | hangs with nothing to interrupt it — the loop only checks its deadline between turns |
| **`getMcpTools`** | one server briefly unreachable at startup | its tools absent from **every later call**, no recovery short of a manual `resetMcpTools()` |
| **`resolveAdapterDefaultModel`** | one failed lookup | that adapter's displayed model blank for the whole process |

The request deadline is the caller's `timeoutMs`, **not a short network timeout**: it covers the streamed body, and anything shorter would cut off legitimate long generations. A complete MCP discovery is still cached indefinitely; only an incomplete one gets a 60s lease, so a blip heals itself without hammering a dead server. A successful model lookup is still cached, since `getDefaultModel` is an OAuth plus catalog round trip.

## Review found two gaps — both fixed here

- The helper exposing which servers were unreachable **had no production consumer**. It is private again. An exported function nothing calls is exactly what the previous PR deleted 962 lines over; adding one in the same session would be poor form.
- The tests exercised `initMcpTools`, leaving the actual cache decision in `getMcpTools` unverified. They now drive `getMcpTools` and go red when the TTL is removed.

## Verification

| Mutation | Tests that went red |
|---|---|
| cache the model lookup failure again | `retries after a failed lookup instead of caching the failure` |
| `getMcpTools` caches forever | `re-discovers once the short lease expires`, `stops re-discovering once a discovery finally completes` |
| deadline ignored | `aborts on its own deadline when the caller never does` |

The MCP tests also pin the **opposite** direction — a complete discovery must not be re-run, and a recovered server must stop being re-discovered — so "always re-discover" is not a way to make them pass.

- Full suite: **256 files, 3402 passed, 1 skipped**
- `openswarm review`: **APPROVE**, 0 issues (after one REVISE round, addressed above)

Named `abortSignalWithDeadline` rather than `withDeadline`: `mcpClient.ts` already exports a `withDeadline` that wraps a Promise, and two functions of that name with different signatures is how a wrong import gets written later.

---

> [!NOTE]
> **Medium Risk**
> Touches long-running worker/provider paths (fetch abort, MCP routing cache, retry classification); behavior is well-tested but incorrect deadline or cache logic could affect all adapter runs and MCP tools in the daemon.
> 
> **Overview**
> Stops three **transient blips from sticking for the process lifetime**: silent provider streams, MCP discovery gaps, and failed default-model lookups.
> 
> **Provider requests** now use `abortSignalWithDeadline` on Atlas Cloud and Codex Responses `fetch` calls (default `timeoutMs` 300000), so a single turn—including a stalled stream—can abort instead of hanging until the agentic loop’s between-turn deadline. **`isInfraError`** treats `DOMException` `TimeoutError` as infrastructure (not a task failure); user `AbortError` stays excluded.
> 
> **`getMcpTools`** keeps complete discoveries cached forever but gives **incomplete** ones a 60s lease (from finish time), dedupes concurrent discovery, keeps prior routing during rediscovery, and ignores stale results after `resetMcpTools` via a generation counter. **`resolveAdapterDefaultModel`** removes failed lookups from the cache so a later call retries while successes stay cached.
> 
> Regression tests cover deadlines, model-cache behavior, and MCP cache/TTL/concurrency/routing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 332cf707a18584ef4c9161ced9d958b330a41ad3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/adapters/atlascloud.ts
+++ b/src/adapters/atlascloud.ts
@@ -11,6 +11,7 @@ import type {
   WorkerResult,
   ReviewResult,
 } from './types.js';
+import { abortSignalWithDeadline } from './requestDeadline.js';
 import {
   runAgenticLoop,
   loopResultToCliResult,
@@ -72,6 +73,7 @@ export class AtlasCloudCliAdapter implements CliAdapter {
     const callApi = createApiCaller(apiKey, model, {
       onToken: options.onToken,
       signal: options.signal,
+      timeoutMs: options.timeoutMs ?? 300000,
     });
 
     const mcpTools = await resolveMcpTools(options.mcpTools);
@@ -127,6 +129,15 @@ export class AtlasCloudCliAdapter implements CliAdapter {
 export interface AtlasCloudApiCallerOptions {
   onToken?: (delta: string) => void;
   signal?: AbortSignal;
+  /**
+   * Ceiling for one API call, including the streamed body.
+   *
+   * Without it the request was bounded only by the caller's signal, and the
+   * agentic loop checks its deadline between turns — so a connection that
+   * accepted the request and then went silent hung until something else
+   * intervened, which for a background worker is never.
+   */
+  timeoutMs?: number;
 }
 
 export function createApiCaller(apiKey: string, model: string, opts: AtlasCloudApiCallerOptions = {}) {
@@ -153,7 +164,8 @@ export function createApiCaller(apiKey: string, model: string, opts: AtlasCloudA
           'Content-Type': 'application/json',
         },
         body: JSON.stringify(body),
-        signal: opts.signal,
+        // The caller's signal AND this call's own deadline. Either one aborts.
+        signal: abortSignalWithDeadline(opts.signal, opts.timeoutMs),
       });
 
       if (!res.ok) {
--- a/src/adapters/codexResponses.ts
+++ b/src/adapters/codexResponses.ts
@@ -13,6 +13,7 @@ import type {
   WorkerResult,
   ReviewResult,
 } from './types.js';
+import { abortSignalWithDeadline } from './requestDeadline.js';
 import { AuthProfileStore, ensureValidToken } from '../auth/index.js';
 import { runAgenticLoop, loopResultToCliResult, type ChatMessage, type AgenticLoopOptions } from './agenticLoop.js';
 import { parseWorkerResult, parseReviewerResult } from './resultParsing.js';
@@ -360,6 +361,7 @@ export class CodexResponsesAdapter implements CliAdapter {
     const cacheKey = `osw-${options.processContext?.taskId ?? 'cli'}-${options.processContext?.stage ?? 'run'}-${model}`;
     const callApi = this.createApiCaller(
       accessToken, accountId, store, model, options.onToken, options.signal, onReasoning, options.disableReasoning, options.reasoningEffort, cacheKey,
+      options.timeoutMs ?? 300000,
     );
 
     const loopOptions: AgenticLoopOptions = {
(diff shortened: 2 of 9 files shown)


unohee:
**Lease expiring early — correct, fixed in 31c49fe.**

The deadline came from a timestamp taken before the `await`, and unreachable servers are precisely the ones that take a long time to fail, so a slow discovery set a deadline already in the past. The lease failed in exactly the case it exists for. Now computed from `Date.now()` after discovery completes, with a test that advances the mocked clock *during* discovery — it goes red against the previous version.

---

**Deadline abort labelled a user stop — measured, and it does not happen.**

The concern would hold if a deadline abort surfaced as `AbortError`, since `agenticLoop.ts:316` checks `err.name === 'AbortError'`. It does not. `AbortSignal.timeout()` aborts with a `TimeoutError`, and `AbortSignal.any` propagates the reason of whichever signal fired:

```
timeout only        : name=TimeoutError  message=The operation was aborted due to timeout
any([user, timeout]): name=TimeoutError  message=The operation was aborted due to timeout
user abort          : name=AbortError    message=This operation was aborted
```

(Probed against a real `fetch` to a server that never responds, on the Node version this package requires.)

So a deadline abort falls through to the ordinary error path and is reported as a failure — which is what it is — while a genuine user abort still produces `AbortError` and is still classified as a stop. No change made, because the distinction the loop relies on already holds.

unohee:
All four fixed in 332cf70, and three of them were defects my own change introduced.

**`isInfraError` and the deadline abort — correct, and my earlier refutation was about a different layer.** I measured it this time:

```
DOMException TimeoutError    isInfraError=false  isTimeoutError=true
AbortError (user)            isInfraError=false  isTimeoutError=false
connect ETIMEDOUT ...        isInfraError=true
```

So adding the deadline turned "the provider accepted the request and went silent" into "the task failed" — a different verdict with a different retry policy. `isInfraError` now recognises `DOMException('TimeoutError')`, matching the precedent `isTimeoutError` already set. A user `AbortError` is deliberately still excluded: that is a stop, not a fault.

(The earlier round's claim was that a deadline abort would be logged as *user cancellation*. That one I measured and refuted — the loop checks `AbortError`, and a timeout produces `TimeoutError`. This finding is a real one at a different layer, and I should have looked further down the path the first time.)

**Discovery failure state shared across concurrent runs — correct.** The list lived on a module variable that `initMcpTools` cleared on entry, so two overlapping discoveries erased each other's records and a partial result could be cached as complete. That recreates the exact process-lifetime tool loss this change exists to prevent. Discovery is now a pure function returning its own `defs`, `routing` and `unreachable`.

**Routing cleared during rediscovery — correct.** In-flight agents saw `MCP tool not registered` for tools that worked a moment earlier. The replacement map is built locally and published only when the run completes.

**No in-flight deduplication — correct.** Concurrent callers each started a discovery, raced to publish, and hit every unreachable server once per caller. One run is now shared.

**Lease measured before the await** was already fixed in 31c49fe.

Also added the reviewer's non-blocking note: a generation counter so a discovery that started before `resetMcpTools` cannot publish afterwards.

Each is pinned by a test that goes red when the specific fix is removed — including the opposite directions, so "always rediscover" or "never cache" cannot satisfy them. Full suite 256 files / 3409 passed.

PR #361, opened 2026-08-01
fix(test): stop the sandbox Git tests timing out under suite load

unohee:
Closes INT-3104.

## Diagnosis

Two tests in `verify/runner.test.ts` intermittently failed with `headStatus: 'infra'` when the whole suite ran, while passing 33/33 in isolation. `'infra'` reads as *the sandbox is broken*, so the failure looked like a real defect in the verify path rather than a flaky test.

It was a timeout. `errorClassification` treats `"timeout after"` as infrastructure, so exceeding the per-command ceiling reports `'infra'` rather than `'fail'` — **a slow sandbox and a broken one produce the same status.**

## Why these two and not the others

They are the only tests in the file that run real Git *inside* the sandbox, which makes them far heavier than their neighbours:

| test | isolated | ceiling | margin |
|---|---|---|---|
| preserves Git metadata | 1350ms | 5s | 3.6x |
| does not share the Git index | 1394ms | 5s | 3.6x |
| mirrors tracked deletions | 196ms | 2s | 10x |
| keeps a failure non-blocking | 297ms | 2s | 6.7x |

Under artificial load (8 spinners on a 10-core machine) the same two rose to **2027ms and 2153ms** — a 1.5x stretch from CPU contention alone, and the full suite runs nine workers on top of that.

## Fix

The ceiling exists to catch a hang, not to assert speed, so it is now wide enough (30s) that only a genuine hang trips it. Both status assertions also carry `rawOutputTail`, so the next `'infra'` failure states whether the sandbox broke or merely ran out of time instead of leaving that to be guessed.

## Verification

- isolated: 33/33 ✅
- **under 8-process CPU load: 33/33** ✅ (this is the condition that used to fail)
- full suite: 3487 passed / 1 skipped (262 files) ✅ · lint ✅

Note the original flake did not reproduce on three clean full-suite runs today — machine conditions differ from when it was filed. The causal mechanism is confirmed by measurement rather than by catching it live, and the load experiment above reproduces the stretch that caused it.

### DIFF
--- a/src/verify/runner.test.ts
+++ b/src/verify/runner.test.ts
@@ -278,9 +278,16 @@ describe('runVerify', () => {
   // under full-suite parallel load they intermittently blew through it. A timeout
   // surfaces as headStatus 'infra' — errorClassification treats "timeout after" as
   // infrastructure — so the failure read as a broken sandbox rather than a slow
-  // one. The ceiling is a hang guard, not a performance assertion, so it is set
-  // wide enough that only a genuine hang trips it. (INT-3104)
-  const SANDBOX_GIT_TIMEOUT_MS = 30_000;
+  // one. The ceiling is a hang guard, not a performance assertion.
+  //
+  // It must stay clearly BELOW vitest's own testTimeout (30s, vitest.config.ts).
+  // runVerify starts this timer only after building the sandbox, so a ceiling at
+  // the harness limit would let vitest kill the test first and report a generic
+  // suite timeout — losing the headStatus and rawOutputTail this guard exists to
+  // produce, which is the same unexplained failure in different clothing. 15s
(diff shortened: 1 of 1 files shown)


unohee:
Good catch — verified and fixed in `1d9bfe4`.

`vitest.config.ts:120` does set `testTimeout: 30000`, and since `runVerify` starts the command timer only after building the sandbox, a command approaching the ceiling would have been killed by vitest first. That would have reported a generic suite timeout and lost the `headStatus` and `rawOutputTail` the guard exists to produce — reintroducing the unexplained failure this change was meant to remove.

Ceiling lowered to **15s**: ~7x the 2.1s measured under artificial load, leaving half of vitest's budget for sandbox setup and teardown. Re-verified under 8 spinners on a 10-core machine — 2551ms and 2111ms, 33/33 passing.

PR #362, opened 2026-08-01
feat(ci): machine-readable review output — JSON and SARIF

unohee:
Part of INT-3102.

`openswarm review` only ever spoke to humans: coloured terminal text plus Linear issues. A CI gate needs the same verdict in a shape other tools can consume.

## `--json`

Emits the verdict on stdout with a versioned schema (`schemaVersion: 1`), **replacing** the human report rather than adding to it — mixing prose into stdout would break `openswarm review --json | jq`. Everything else the command says already routes through `log`.

```json
{
  "schemaVersion": 1,
  "decision": "revise",
  "gateRan": true,
  "feedback": "...",
  "issues": [], "suggestions": [],
  "findings": [
    { "type": "bug", "title": "Restore CSRF validation",
      "location": "src/auth/session.ts:42", "file": "src/auth/session.ts", "line": 42 }
  ]
}
```

## `--sarif <file>`

SARIF 2.1.0, so GitHub code scanning can surface findings inline on a PR.

## Two judgement calls

**Reviewer `location` is free-form prose in practice** — `src/foo.ts:42`, `src/foo.ts`, or `the auth middleware`. Only the first two can be placed in a file, so `parseLocation` refuses anything that does not look like a path. A finding whose location is prose still appears in the report but carries no `physicalLocation`: dropping it would silently shrink the report, and inventing a file would annotate a path that does not exist. The line is taken only from a trailing all-digits segment so a Windows drive letter (`C:\src\foo.ts`) is not read as one.

**SARIF findings are `warning`, not `error`.** The blocking decision is the verdict; individual follow-ups are advisory and several are emitted alongside an `approve`. Reporting them as errors would fail code scanning on suggestions the review itself accepted.

A SARIF file that cannot be written is a warning, not a failed gate — the verdict already exists and still decides the exit code (INT-3100).

## Not in this PR

PR annotations, the third leg of INT-3102, land with the GitHub Action in INT-3101 — the annotation has to be emitted from inside the workflow.

## Gates

`lint` ✅ · `typecheck` ✅ · `test` **3502 passed / 1 skipped (263 files)** ✅ · 15 new tests

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -319,6 +319,15 @@ program
     out?: string; linear?: boolean; fallback?: string | boolean; fix?: boolean; inPlace?: boolean; fixRounds?: number; learn?: boolean;
   }) => {
     try {
+      // --json/--sarif are declared on the shared `review` command but only the
+      // direct path implements them. Silently printing the normal report and
+      // never writing the requested file is worse than refusing: a CI step would
+      // read an absent SARIF as "no findings".
+      if (opts.max && (opts.json || opts.sarif)) {
+        console.error('--json and --sarif are not supported with --max yet; use --out for the audit report.');
+        process.exitCode = 2;
+        return;
+      }
       if (opts.max) {
         const { runReviewMaxCommand, reviewMaxResultFailed } = await import('./cli/reviewMaxCommand.js');
         const result = await runReviewMaxCommand({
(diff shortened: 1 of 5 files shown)


unohee:
All six findings were real and are fixed in `51991d5` + `6cbd0a5`. Thanks — the two P1s defeated the feature's purpose outright.

**P1 · JSON mode wrote prose to stdout.** The comment claimed everything else routed to stderr; nothing implemented that. `log` defaulted to `console.log`, and in CI `process.stderr.isTTY` is false so reviewer progress took the same path. `log` now writes to stderr whenever `--json` is set.

**P1 · SARIF dropped blocking issues.** Issues are emitted at `error`, follow-ups stay at `warning` — the split the reviewer prompt already makes.

**P2 · `--json`/`--sarif` with `--max`** now fails explicitly instead of silently printing the normal report; an absent SARIF would have read as "no findings".

**P2 · `artifactLocation.uri`** is now a proper URI reference — separators normalized, segments percent-encoded.

**P2 · extensionless root files** keep their location via a known-names set. Worth noting: the first attempt at this rejected all whitespace, which broke legitimate paths containing spaces — caught by the URI test in the same batch. A separator or dotted suffix is now decisive on its own, and the whitespace test applies only to bare words.

**P2 · `gateRan: false`** is now reachable — the failure path writes the documented document with an empty decision rather than inventing a verdict.

Gates: lint · typecheck · **3515 passed / 1 skipped**, coverage 90.05%. 21 new tests.

PR #364, opened 2026-08-01
feat(ci): package the review gate as a GitHub Action

unohee:
Part of INT-3101.

`openswarm review` already had the exit-code contract a merge gate needs (INT-3100) and, since #362, machine-readable output. What was missing was the wrapper: every consumer had to hand-write the install, the base fetch, the flag set, and the exit-code mapping.

## Two details a hand-rolled step usually gets wrong

**The base branch isn't there.** `actions/checkout` fetches only the PR head, so `git diff` against the base finds nothing and **the gate passes vacuously** — the worst possible failure for a merge gate. The action fetches the base ref first.

**`exit 2` must fail.** That is the entire point of INT-3100's 1/2 split: a review that did not happen must not read as a pass. Default is to fail the job; `fail-on-gate-not-run: false` exists for callers with a separate alert, and it is opt-in.

## Usage

```yaml
permissions:
  contents: read
  security-events: write   # only for the SARIF upload

steps:
  - uses: actions/checkout@v4
    with: { fetch-depth: 0 }
  - uses: actions/setup-node@v4
    with: { node-version: '22' }
  - id: review
    uses: unohee/OpenSwarm@main
    env:
      OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
  - if: always() && steps.review.outputs.sarif-file != ''
    uses: github/codeql-action/upload-sarif@v3
    with: { sarif_file: ${{ steps.review.outputs.sarif-file }} }
```

Inputs: `base`, `adapter`, `version`, `sarif-file`, `fail-on-gate-not-run`.
Outputs: `decision`, `gate-ran`, `sarif-file`.

SARIF upload stays with the caller because it needs `security-events: write`, which a composite action cannot grant itself.

## The dogfood workflow is manual on purpose

`.github/workflows/review-gate.yml` is `workflow_dispatch`, **not** `on: pull_request`. The gate calls a paid model on every run, so enabling it for every PR is a spending decision the repository owner should make deliberately rather than inherit from a merge. It pins routing to the sponsor's capacity — without that the account's own credit is spent, and it has none.

## Verification

Both YAML files parse, and the composite step's shell script passes `bash -n`. **The action itself has not been executed** — that needs a dispatch on a real PR with the secrets configured, which is the natural next step once this is on `main`.

## Remaining on INT-3101

Token-only auth for Linear (adapter keys already read from env) and the CLI-install path for `codex`/`claude` adapters in a container. Neither blocks the action for API-direct adapters.

### DIFF
--- a/.github/workflows/review-gate.yml
+++ b/.github/workflows/review-gate.yml
@@ -5,7 +5,14 @@ name: Review Gate
 # Deliberately manual (workflow_dispatch), not `on: pull_request`. The gate calls
 # a paid model on every run, so turning it on for every PR is a spending decision
 # the repository owner should make explicitly rather than inherit from a merge.
-# To enable it, add the `pull_request` trigger below.
+#
+# To automate it, the trigger you want is `pull_request_target`, not
+# `pull_request`: secrets are not exposed to a `pull_request` run from a fork, so
+# OPENROUTER_API_KEY would be empty and every fork PR would fail as gate-not-run.
+# `pull_request_target` runs with secrets — which is safe here only because the
+# two checkouts below keep the executed action code and the reviewed code apart,
+# and the reviewer itself runs read-only. The PR number resolves from either
+# trigger already.
 
 on:
   workflow_dispatch:
@@ -27,10 +34,45 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
-      - uses: actions/checkout@v4
+      # The action's own code comes from the default branch, never from the pull
+      # request. `uses: ./` against a PR checkout executes action.yml as the
+      # contributor wrote it, and this job hands that code OPENROUTER_API_KEY —
+      # so a PR that edits action.yml would be a credential handout. Two
+      # checkouts in two paths keep the executed code and the reviewed code
+      # apart.
+      #
+      # Consequence worth knowing: this workflow cannot run against the pull
+      # request that introduces action.yml, because the default branch does not
+      # have that file yet. The first real execution is necessarily after merge.
+      - name: Check out the trusted action code
+        uses: actions/checkout@v4
         with:
-          ref: refs/pull/${{ inputs.pr }}/head
-          # The gate diffs against the base branch, which a shallow single-ref
+          ref: ${{ github.event.repository.default_branch }}
+          path: action-src
+
+      - name: Resolve the pull request number
+        id: pr
+        env:
+          # From the event when a PR triggered this, from the dispatch input
+          # otherwise. Reading only `inputs.pr` would produce `refs/pull//head`
+          # the moment the `pull_request` trigger above is enabled.
+          EVENT_PR: ${{ github.event.pull_request.number }}
+          INPUT_PR: ${{ inputs.pr }}
+        run: |
+          set -euo pipefail
+          NUMBER="${EVENT_PR:-$INPUT_PR}"
+          if [ -z "$NUMBER" ]; then
+            echo "::error::No pull request number from the event or the dispatch input"
+            exit 1
+          fi
+          echo "number=$NUMBER" >> "$GITHUB_OUTPUT"
+
+      - name: Check out the pull request
+        uses: actions/checkout@v4
+        with:
+          ref: refs/pull/${{ steps.pr.outputs.number }}/head
+          path: pr
+          # The gate diffs against the merge base, which a shallow single-ref
           # clone does not contain.
           fetch-depth: 0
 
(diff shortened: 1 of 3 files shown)


unohee:
Converting to draft. The review found 8 issues and **two are security blockers I cannot fix inside this PR**.

## Blocker 1 — the reviewer runs untrusted code with credentials present

`runReviewer` calls `spawnCli` without `enableTools: false` or any read-only flag (`src/agents/reviewer.ts:257`), and the adapters default to `enableTools ?? true`. **The reviewer therefore has bash.** In this action that means an agent with shell access running over attacker-controlled PR files while `OPENROUTER_API_KEY` is in the environment — a prompt-injection path straight to credential exfiltration.

`ReviewerOptions` has no `readOnly` field at all, so this is not a flag I can pass; the reviewer needs a read-only mode first. Filed as a blocker.

## Blocker 2 — `uses: ./` executes action code from the PR

The workflow checks out `refs/pull/<n>/head` and then runs `uses: ./`, so `action.yml` comes from the untrusted PR rather than the trusted default branch. A contributor can rewrite it to do anything with the secrets the job holds.

## The other six, all correct

- `--json`/`--sarif` did not exist at this commit — they landed in #362, merged since, so a rebase fixes it
- `|| true` on the base fetch means a failed fetch reviews nothing and **passes vacuously** — the exact failure this gate exists to prevent
- no adapter is selected, so the CLI falls back to its default while OpenRouter credentials are exposed
- should diff against the **merge base**, not the base tip, or unrelated target-branch commits enter the diff
- `review.json` is created inside the checkout and gets picked up as a changed file
- the README's "just add a `pull_request` trigger" does not populate `inputs.pr`

## Plan

The read-only reviewer lands first; this PR rebases onto it and fixes the rest. Shipping the action without it would hand a credential-exfiltration path to anyone who can open a pull request.

unohee:
Fix pass pushed (7727fd5), rebased onto main. All 8 original findings addressed, and a second independent review of the fix pass found 8 more — those are handled here too, except one that became its own PR.

## Original 8

| # | Status |
| --- | --- |
| `--json` / `--sarif` unsupported | Landed separately (INT-3102), and the action now **preflights** `review --help` for every flag it passes |
| base fetch `\|\| true` | Removed — a failed fetch exits 2 without invoking the reviewer |
| `uses: ./` on untrusted checkout | Two checkouts: action code from the default branch, PR into `pr/`, `path: pr` |
| no adapter selected | `adapter: openrouter` passed explicitly |
| two-dot diff vs base tip | Merge base, in both the action and the workflow |
| reviewer had bash | `--read-only` default true (INT-3189), plus #367 below |
| `review.json` inside the checkout | Both scratch files under `$RUNNER_TEMP` |
| PR number from `inputs.pr` only | `${EVENT_PR:-$INPUT_PR}` |

## Found in the fix pass

**Exit 1 meant "rejected" even when nothing was reviewed.** npm `latest` is 0.19.12, whose published `dist/cli.js` has neither `--sarif` nor `--read-only` — I verified this against the actual tarball. Commander exits 1 on an unknown option, and 1 is the reject code, so as written this action would have reported *every* run as a rejected change. Two fixes: a capability preflight that fails as gate-not-run, and a rule that a non-zero exit is only a reject when a verdict was actually parsed.

**`echo "ref=$(git merge-base …)"` escaped `set -e`.** A command substitution that fails inside an argument does not trip `set -e` — confirmed at a shell. The step succeeded with an empty ref, the action then got no `--base`, reviewed a clean working tree, found nothing, and went green. Now assigned to a variable first.

**A `config.yaml` in the reviewed PR could execute arbitrary code with the provider key.** `loadConfig` searches cwd first, `mcp.servers` from that file gets merged into the registry, and resolving MCP tools *spawns* each server with `${OPENROUTER_API_KEY}` expanded into its environment — before the read-only filter, which only hides tools from the model, ever runs. The reviewer that found it had a working PoC. Two layers of fix: **#367** stops read-only runs from resolving MCP at all, and this action no longer makes the checkout its working directory (`--path` instead), so the config is never discovered.

**Boolean inputs were fail-open.** `[ "$X" = "true" ]` turned `True` into a silently disabled `read-only`. Both booleans are now parsed and reject anything else.

**SARIF upload was attributed to the dispatching ref**, not the analysed commit — `ref` and `sha` are passed now.

**Docs corrected**: the two-checkout dance is only needed for `uses: ./`; external callers using `unohee/OpenSwarm@main` already get trusted code. And automation needs `pull_request_target`, since a fork's `pull_request` run has no secrets and would fail every PR as gate-not-run.

## Known limitation

This workflow cannot run against this PR — the default branch has no `action.yml` until it merges, so the trusted checkout finds nothing. Stated in the workflow comment rather than left as a surprise. In its place the shell was exercised against a local git fixture: merge base versus a moved base branch, unknown base, old CLI, verdict-less crash, bad boolean, absolute path, and each arm of the 0/1/2 exit contract.

Blocked on #367 landing first.

PR #365, opened 2026-08-01
fix(verify): probe bwrap instead of guessing from sysctls

unohee:
Follow-up to #363, which I merged before its review landed. All six findings were right, and the deepest one invalidated the approach.

## Reading sysctls answers the wrong question

It was wrong in **both** directions:

- **False pass** — a container can deny namespace creation through seccomp, dropped capabilities, or `user.max_user_namespaces=0` while every sysctl checked reads permissive. The check passed and the run died later at exec time, which is exactly the failure #363 set out to remove.
- **False fail** — `unprivileged_userns_clone=0` does not stop a setuid-root bwrap or a caller holding `CAP_SYS_ADMIN`, and the AppArmor sysctl means userns is *mediated*, not denied: a profile can permit it. Working hosts were rejected outright.

Availability is now decided by running `bwrap --ro-bind / / --unshare-user true`. The sysctls are still read — but only afterwards, to explain a denial that already happened.

## The remedies were wrong too

`seccomp=unconfined` does not change a kernel sysctl and does not alter AppArmor mediation, so offering it under either cause sent the operator after a fix that cannot work. It now appears only on the generic container case, alongside `--cap-add SYS_ADMIN` and `user.max_user_namespaces`. The `apt-get` line gained the `sudo` that `ubuntu-latest` needs, since workflow steps do not run as root.

## Two details

The probe result is **cached per process** — it cannot change mid-run, and every verification command would otherwise pay for another namespace creation.

The wiring is a pure factory (`makeSystemProbe`) rather than inline at the call site: the Linux branch never executes on macOS, so anything assembled there ships untested. That is not hypothetical — the first version of this file was merged with its whole wiring uncovered, and the coverage gate caught it here.

## Gates

`lint` ✅ · `typecheck` ✅ · `test:coverage` **3501 passed / 1 skipped, 90.02%** ✅ · 12 tests on this module

### DIFF
--- a/src/verify/runner.ts
+++ b/src/verify/runner.ts
@@ -1,4 +1,4 @@
-import { execFile, spawn } from 'node:child_process';
+import { execFile, spawn, spawnSync } from 'node:child_process';
 import { createHash, randomUUID } from 'node:crypto';
 import { access, cp, lstat, mkdir, mkdtemp, readFile, readdir, readlink, realpath, rm } from 'node:fs/promises';
 import { existsSync, readFileSync } from 'node:fs';
@@ -6,7 +6,7 @@ import { tmpdir } from 'node:os';
 import { delimiter, dirname, isAbsolute, join, relative, resolve, sep } from 'node:path';
 import { promisify } from 'node:util';
 import { isInfraError } from '../adapters/errorClassification.js';
-import { describeLinuxSandbox, formatSandboxUnavailable } from './sandboxDiagnostics.js';
+import { describeLinuxSandbox, formatSandboxUnavailable, makeSystemProbe } from './sandboxDiagnostics.js';
 import { copyIsolatedPath } from '../support/isolatedPath.js';
 import { loadRepoMetadata } from '../support/repoMetadata.js';
 import { resolveSharedPaths } from '../support/worktreeManager.js';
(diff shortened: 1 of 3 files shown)


unohee:
All five findings addressed in 18fa803 — every one was real.

**P1 · probe the namespaces the runner uses.** The probe exercised `--unshare-user` only; the real invocation also sets up `--unshare-net --dev /dev --proc /proc`. A host permitting user namespaces but denying network ones passed the probe and then failed every verification command — the exact failure this module exists to catch. The probe now mirrors the runner's argv, and a test *reads* that argv rather than restating it, so the two cannot drift apart silently.

**P1 · absolute probe command.** `true` was resolved with execvp against the inherited PATH inside the sandbox, so a project-supplied `node_modules/.bin` decided what ran. Now `/usr/bin/true` or `/bin/true` by absolute path, falling back to bwrap itself (known to exist by that point).

**P2 · re-probe after remediation.** Only a working sandbox is memoized now; a broken one is re-probed, so an operator who installs bubblewrap or flips the sysctl is unblocked without restarting the daemon. Cost is one `bwrap true` on a host that is already broken. The decision moved into `makeSandboxCache` so it is testable without a Linux host.

**P2 · AppArmor attribution.** Mediation being on is not proof it denied us. It is now named as the *likeliest* cause, the generic remedies stay attached for when it was not, and bwrap's own stderr is quoted in every reason — including the sysctl-attributed ones — so the operator sees ground truth instead of a guess.

**P2 · root containers.** Dropped `sudo` from the package-install remedies: the environment where bwrap is most often missing is a minimal Debian/Alpine container running as root with no sudo, where the advertised command fails before reaching the package manager. The GitHub Actions line keeps `sudo`, where the runner is specifically not root.

Gate: lint clean, typecheck clean, 3508 tests pass, statements 90.04%.

PR #367, opened 2026-08-01
fix(adapters): a read-only run must not connect MCP servers (INT-3189)

unohee:
## The hole

The read-only reviewer denied `bash` and the write tools — and still had arbitrary code execution, because **resolving MCP tools means connecting to them**.

Every step of the chain lives in the repository under review:

1. `cli.ts` preAction `loadConfig()` searches `process.cwd()` **first** (`src/core/config.ts:25-31`), and during a CI review the cwd is the checked-out pull request.
2. A `config.yaml` there is accepted — `looksLikeOpenSwarmConfig` only requires a top-level `agents:`.
3. `substituteEnvVarsDeep` expands `${OPENROUTER_API_KEY}` to its real value.
4. `mcpClient` merges that file's `mcp.servers` into the registry.
5. `resolveMcpTools` opens a `StdioClientTransport` per server — which **spawns the configured command with the configured env**.

All four API adapters called `resolveMcpTools` unconditionally, before the loop's read-only filter (`agenticLoop.ts`, `readOnly ? [] : mcpTools`). That filter only removes the tools from the list the model is *shown*. By then the attacker's command has run and been handed the credential. A failed handshake does not save you either — the spawn happens first, and `server unreachable — skipped` is printed after.

## The fix

Skip at the resolve, not at the filter. A read-only run contacts no MCP server at all, which costs nothing: the loop was going to discard those tools anyway.

## Verification

`src/adapters/readOnlyMcp.test.ts` asserts, for each of the four adapters, that `resolveMcpTools` is **not called** under `readOnly` and *is* called for an ordinary run. Confirmed the pair bisects: reverting the openrouter line alone turns that adapter's case red.

- `npm run lint` · `npx tsc --noEmit` — clean
- `npm run test:coverage` — 265 files, 3557 pass, statements 90.15%

## Credit

Found by an independent reviewer while reviewing the GitHub Action PR (#364), with a working proof of concept: a `config.yaml` in the repo root, no flags, and the marker file existed with the key inside it. #364 stays draft until this lands.

### DIFF
--- a/src/adapters/atlascloud.ts
+++ b/src/adapters/atlascloud.ts
@@ -120,7 +120,17 @@ export class AtlasCloudCliAdapter implements CliAdapter {
       timeoutMs: options.timeoutMs ?? 300000,
     });
 
-    const mcpTools = await resolveMcpTools(options.mcpTools);
+    // MCP tools: caller-provided, else self-source from the registry. (INT-1951)
+    //
+    // Skipped entirely in read-only mode, and the skip has to be here rather
+    // than in the loop. Resolving means CONNECTING: the registry merges
+    // `mcp.servers` from the config discovered in cwd, and the stdio transport
+    // spawns each server's command with `${SECRETS}` already expanded into its
+    // environment. When the cwd is a checkout under review, that config is
+    // attacker-authored, so a read-only run that resolved first and filtered
+    // afterwards would have executed the attacker's command and handed it the
+    // provider credential before any tool list was consulted. (INT-3189)
+    const mcpTools = options.readOnly ? undefined : await resolveMcpTools(options.mcpTools);
 
     const loopOptions: AgenticLoopOptions = {
       systemPrompt: options.systemPrompt,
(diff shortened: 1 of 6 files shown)


unohee:
All three findings addressed.

**P1 · read-only denied MCP discovery but not MCP calls** — the important one, and a real gap in my fix. Skipping discovery closes the door for a fresh process, not for the daemon: a long-lived run that resolved those servers during an ordinary task still has them in `serverByTool`, so a later read-only run whose model emits `server__tool` reaches one. The executor already exists precisely because a model calls tools it was never shown. Denied there too now, **by predicate rather than by name** — MCP names are whatever the servers declare, and a fixed list that looked complete would be worse than none.

**P1 ×2 · the `gpt` and `local` test rows never reached the code under test** — correct, and CI proved it before this review landed: both passed on my machine only because it has a live LM Studio and OAuth creds, and failed on the runner. Rather than mock an OAuth profile and an endpoint probe, the behavioural pair now covers the two adapters reachable from an API key alone, and the other two are covered by a source invariant — every `resolveMcpTools` call must be guarded on `readOnly`. That also catches an adapter added later, which no per-adapter list would.

Each new test was checked to bisect: reverting the corresponding line turns exactly one case red, for its own reason.

Gate: lint clean, typecheck clean, 3559 tests pass, statements 90.14%.

PR #369, opened 2026-08-01
ci: measure whether the verify sandbox works on ubuntu-latest (INT-3103)

unohee:
The README tells operators that deterministic verification needs bubblewrap plus unprivileged user namespaces, and that on GitHub's runners installing the package is enough. Nothing checked that claim, and no unit test can — it is a property of the runner, not of our code.

This job probes before and after installing, and uses **the same namespace setup `src/verify/runner.ts` does** (`--unshare-net --dev /dev --proc /proc`). Probing a subset would pass on a host that denies one of the others, which is precisely the failure #365 was about.

Closes the measurement item in INT-3103's DoD.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,41 +112,34 @@ jobs:
           done
 
   verify-sandbox:
-    # Deterministic verification refuses to run without an OS sandbox, and the
-    # answer depends on the runner, not on our code: bubblewrap is absent from
-    # ubuntu-latest by default, and even installed it needs unprivileged user
-    # namespaces the host may deny. Measuring it here is the only way that claim
-    # in the README stays true — a unit test cannot tell you what this runner
-    # allows. (INT-3103)
+    # Deterministic verification refuses to run without an OS sandbox, and
+    # whether one is available is a property of the runner, not of our code — no
+    # unit test can answer it. This job asserts the recipe the README gives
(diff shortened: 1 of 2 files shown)


unohee:
Correct against the first commit, and already fixed by the two after it — the finding is the whole point of the PR rather than an objection to it.

The hard probe without the sysctl is exactly what ran, and it failed on the runner with `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`. Turning the job into a diagnostic matrix produced the real picture on `ubuntu-latest`:

```
kernel.unprivileged_userns_clone=1
kernel.apparmor_restrict_unprivileged_userns=1
user.max_user_namespaces=63838
  FAIL  user namespace only — bwrap: setting up uid map: Permission denied
  FAIL  + /proc + /dev        — bwrap: setting up uid map: Permission denied
  FAIL  + network namespace   — bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted
```

Two permissive sysctls and one restrictive one, and the restrictive one wins — which is a fair summary of why inferring availability from those files was the wrong approach in the first place (#365).

The job now asserts both halves of the claim: that bubblewrap alone does **not** suffice on this runner (warning if a future image changes that, so the docs cannot rot), and that install + `sysctl -w kernel.apparmor_restrict_unprivileged_userns=0` does. Both steps pass. The README gained a per-environment table stating the sysctl, which the Tests job had been doing silently since before this PR.

PR #374, opened 2026-08-01
fix(ci): show why the gate said revise, not just that it did

unohee:
The first run of the gate with working tools returned `revise`. The job log contained the single word "revise".

`--json` suppresses the human report by design, and SARIF carries only findings that have a file and line — the analysis for that run uploaded **0 results**. So the feedback and any unlocated issues, which is often the whole of a `revise`, appeared nowhere. An operator gets a verdict and no way to act on it, which is most of the gate's value missing.

## Fix

The review document is rendered into both the step log and the **GitHub job summary**: decision, feedback, issues, and follow-ups with locations.

The exit-1 message now quotes the actual decision instead of asserting "rejected this change", so it cannot misreport if that mapping ever widens beyond reject.

## Verification

Exercised against the local fixture harness with a `revise` document carrying feedback, an unlocated issue, and two follow-ups (one with a location, one without). All of it reaches the log and `$GITHUB_STEP_SUMMARY`; `bash -n` and YAML parse clean.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -233,12 +233,21 @@ runs:
             out.push(`## OpenSwarm review: ${r.decision || "gate did not run"}`);
             if (r.feedback) out.push("", r.feedback);
             if (r.issues?.length) out.push("", "### Issues", ...r.issues.map((i) => `- ${i}`));
+            if (r.suggestions?.length) out.push("", "### Suggestions", ...r.suggestions.map((x) => `- ${x}`));
             if (r.findings?.length) {
               out.push("", "### Follow-ups");
               for (const f of r.findings) out.push(`- [${f.type}] ${f.title}${f.location ? ` — \`${f.location}\`` : ""}`);
             }
             const text = out.join("\n");
-            process.stdout.write(`${text}\n`);
+            // Every word of this came from a model that just read an untrusted
+            // diff. GitHub turns any stdout line beginning with `::` into a
+            // workflow command, so printing it raw would let injected prose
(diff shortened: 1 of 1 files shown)


unohee:
Both findings addressed — the first was a hole this PR opened, which is a fair catch on a change whose entire subject is making a gate honest.

**P2 · workflow commands in review prose.** Rendering the document to the step log meant printing model-generated text, written after reading an untrusted diff, straight to stdout. Any stdout line starting with `::` is a workflow command, so injected prose could forge annotations or emit `::stop-commands::` and silence the rest of the step — including the verdict. The block is now wrapped in stop-commands with a random token the writer cannot predict.

Verified against a review whose feedback contains both `::error::FORGED ANNOTATION` and `::stop-commands::pwned`:

```
::stop-commands::c215cc54-…
## OpenSwarm review: revise
Looks fine.
::error::FORGED ANNOTATION
::stop-commands::pwned
…
::c215cc54-…::
::notice::OpenSwarm review: revise      ← still executes
```

Both forgeries render as text, and the real `::notice::` after the closing token still works — so the fence closes cleanly rather than leaving commands disabled for the remainder of the step. The job summary is Markdown and is not command-parsed, so it takes the text as-is.

**P2 · dropped suggestions.** Correct, and it defeated the point: `suggestions` is part of the `ReviewJson` contract and the normal report renders it, so a failing `revise` could still withhold the proposed fix. Added as its own section before the follow-ups.

PR #377, opened 2026-08-01
fix(review): a diff limit that lets its own warning be truncated is not a limit

unohee:
Found by the gate reviewing the commit that added it — the first time this loop has closed on itself. Its exact words:

> Consider lowering `maxBytes` in `getDiffText` to match or slightly exceed `MAX_PROMPT_DATA_CHARS` (20KB) so the truncation message in `getDiffText` actually reaches the prompt

It is right, and the consequence is worse than the phrasing suggests.

`getDiffText` bounded the diff at 200KB and appended a notice saying how much it dropped and to read the files for the rest. The template bounds the **whole worker report** as untrusted data at 20,000 characters. For any diff between those numbers the template cut first — and what it cut off was the notice. The reviewer got a diff stopping mid-hunk with a bare `[truncated]` and no idea how much was missing.

So #375's claim that it "says so when it truncates" was **false in exactly the range where it mattered**.

## Fix

Default is 16,000 — under the ceiling, with room for the file list that shares the budget.

Rather than couple to that private constant, the test asserts the **property**: build a prompt from a large diff and require the informative notice to survive into it. Confirmed to bisect — restoring 200KB turns it red. It fails if either limit drifts, in either direction.

Also added the `getDiffText` unit coverage the reviewer asked for: content, base-vs-HEAD (where the HEAD diff is empty and only the base ref shows the change — the committed-diff case the whole feature exists for), truncation, and the error path returning empty so a git failure degrades the review instead of ending it.

- `npm run lint` · `npx tsc --noEmit` — clean
- `npm run test:coverage` — 266 files, 3579 pass, statements 90.14%

### DIFF
--- a/src/agents/reviewer.test.ts
+++ b/src/agents/reviewer.test.ts
@@ -583,3 +583,44 @@ describe('the reviewer is given the diff, not asked to find it (INT-3101)', () =
     expect(prompt).not.toContain('Diff under review');
   });
 });
+
+describe('a truncated diff still tells the reviewer it was truncated (INT-3101)', () => {
+  it('keeps getDiffText\'s notice inside the prompt template\'s own ceiling', async () => {
+    // The template bounds the whole worker report as untrusted data and appends
+    // a bare `[truncated]`. If the diff limit sits above that ceiling, the
+    // template cuts first and the informative notice — the part telling the
+    // reviewer to read the files for the rest — is what gets cut. A limit that
+    // lets its own warning be truncated is not a limit.
+    const { getDiffText } = await import('../support/gitTracker.js');
+    const { mkdtempSync, writeFileSync } = await import('node:fs');
+    const { tmpdir } = await import('node:os');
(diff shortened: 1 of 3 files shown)


unohee:
Correct, and it dismantles the fix rather than refining it. 16,000 does not guarantee the notice survives, because the file summary — up to 20 paths of unbounded length — is prepended before the combined report meets the ceiling. "Under the ceiling with room for the file list" was arithmetic I never actually did.

Picking a smaller number would be the same mistake with a different constant, and it would break again the next time either limit moves.

**The notice now leads the diff.** Downstream truncation removes the *end* of a text, so anything appended after the diff is the first thing lost and anything before it survives every later cut. No reasoning about how our limit compares to anyone else's is required.

The test grew the case you described — 20 paths of ~250 characters ahead of an oversized diff — and it bisects: moving the notice back to the end turns it red.

Gate: lint clean, typecheck clean, 3580 tests pass, statements 90.16%.

PR #404, opened 2026-08-10
fix(net): make the outbound SSRF guard actually work (INT-3513)

unohee:
First batch of the 2026-08-09 `review --max` remediation (INT-3513). It lands the outbound-destination hardening — and repairs it, because as generated it disabled every outbound request in the process.

## What was broken

The remediation built a DNS-pinning undici `Agent` and handed it to Node's global `fetch`. Three independent defects:

| Defect | Evidence |
|---|---|
| `undici` imported but never declared in `package.json` | resolved only via a hoisted transitive 6.28.0 |
| npm-`undici` dispatcher rejected by the undici bundled in Node's global `fetch` | `InvalidArgumentError: invalid onError method`; the same URL over plain `fetch` returns 200 |
| `connect.lookup` answered a single address while undici asked for `all` | `TypeError [ERR_INVALID_IP_ADDRESS]: Invalid IP address: undefined` |

Measured effect: **every** `web_fetch`, all three `web_search` backends (Tavily, Brave, DuckDuckGo), and every outbound notification failed.

A fourth defect was independent of the dispatcher: response bodies over 80 KB **threw** instead of truncating, so `webFetch` failed on any normal page — a 397 KB Wikipedia article returned `Fetch failed for …`.

All 14 unit tests passed the whole time. They stub `global.fetch`, so the dispatcher was never exercised.

## Fixes

- declare `undici`, and send validated requests through undici's own `fetch` so client and dispatcher come from one instance
- collapse the two divergent guard copies (webTools, notifier) into one `publicFetch` in `support/outboundUrl.ts`, also used by `pairWebhook`
- honour the `options.all` callback contract
- reject IPv6 multicast `ff00::/8` and site-local `fec0::/10` — the IPv4 branch already folds these into `a >= 224`
- truncate oversized page bodies; keep erroring for JSON endpoints, where a partial document only yields a confusing parse failure

## Verification

- `tsc --noEmit` clean, `oxlint` 0/0
- full suite: 295 files pass
- new tests drive the lookup contract through an injected resolver — no network, and they fail against the shape that shipped
- **live**: the 397 KB page now returns 20,035 chars of text; `127.0.0.1` and `localhost` are refused

## Scope

Advances INT-3492, INT-3411, INT-3461 — none are closed here, since each also covers files outside this batch.


unohee:
## 🔍 Fresh review of 6135c0f (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Manual redirect handling preserves the original POST method and body for every redirect status, diverging from Fetch semantics for 301/302/303. This can make Tavily search fail or repeat an unintended POST when its endpoint redirects.

unohee:
## 🔍 Fresh review of 4e6c340 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  No material correctness or security defect found in the committed changes. Outbound callers now use DNS-validated, public-only transport; web-tool redirects are bounded, revalidated per hop, and prevent credential/body forwarding across origins.
  Suggestions:
    - `src/support/outboundUrl.test.ts` tests the lookup callback shape but not `publicFetch` with an actual undici dispatcher; add an integration test using a controlled DNS resolver/server if practical.

unohee:
## 🔍 Fresh review of d31d0cc (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Reviewed production wiring and changed tests. `publicFetch` is used by web tools, pair webhooks, and notification transports; it validates destinations before request dispatch and enforces public-only DNS resolution at socket lookup time. Manual redirect handling revalidates each web-tools hop and prevents cross-origin forwarding of request bodies or recognized credential headers.

Issue #29, opened 2026-03-09
GitHub Actions CI/CD pipeline


unohee:
## Summary

Add GitHub Actions workflows for automated testing, linting, and release management.

## Proposed Workflows

- **CI**: Run `vitest`, `oxlint`, `tsc --noEmit` on every PR
- **Release**: Auto-publish GitHub releases on version tag push
- **Docker**: Build and push Docker image on release

## Benefits

- Ensure PR quality before merge
- Automate release process
- Provide CI status badges for README

PR #408, opened 2026-08-21
fix(review): stop fabricating a REVISE verdict the reviewer never gave

unohee:
## Problem

`openswarm pr review --fresh` reported `Decision: REVISE` **nine consecutive times** on PRs whose real conclusion was `approve`, posting the reviewer's *opening narration* ("커밋된 변경사항을 …확인합니다") as the feedback. The gate became unusable for automated judgement — and worse, trained a human to disbelieve a real REVISE.

## Root cause — the parser, not the plan tool

The issue suspected codex's plan tool. A live `codex exec --json --sandbox read-only` probe with the real reviewer prompt showed `update_plan` emits `item.type == "todo_list"` (`{text, completed}` entries) — a shape with nowhere to put a verdict. The actual chain:

1. `extractCodexMessageText` took the **last** `agent_message` as the final output
2. A turn ending without a concluding message leaves the reviewer's **first** utterance there
3. `parseReviewerResult` got text with no JSON and no `Decision:` line
4. `extractReviewerFromText` defaulted the verdict to `revise`
5. `isSubstantiated`'s text path accepted **any residual prose** as evidence → guard passed

The parser was internally inconsistent: it rejected `Decision: revise` (verdict, no findings) while accepting prose with no verdict at all.

## Changes

| Area | Change |
|---|---|
| `adapters/resultParsing.ts` | Refuse a verdict that was never declared. Prose substantiates only when a verdict was explicitly stated; "declared but empty" and "no verdict at all" report differently |
| `adapters/codex.ts` | Scan `agent_message`s **backward** for the last parseable verdict, so a sign-off after it isn't discarded. Non-final messages held to a **JSON-only** bar |
| `automation/prProcessor.ts`, `cli/prCommand.ts` | `gateRan` signal; exit **2** (gate did not run) split from exit **1** (reviewer rejected) |
| `automation/prProcessor.ts`, `cli/reviewHistory.ts` | PR history written to the real repo as `kind: 'pr'` **and read from there** — it previously went to the scratch worktree that `finally` deletes. Kind allow-list unified so the runtime guard and the type can't drift |
| `locale/prompts/{en,ko}.ts` | Reviewer told to emit its JSON as the final message, not via a tool |

### Why non-final messages are JSON-only

Raised in independent review: on a truncated turn, mid-stream prose like *"my decision: approve, pending one final check"* matches the verdict regex. Accepting it would make this merge gate a **silent pass** — strictly worse than the noisy REVISE being fixed. JSON is a shape reviewers only emit when reporting a result.

## Verification

**Against the reported reproduction.** vega-plugins PR #29 rebuilt exactly (single-commit PR, so `743594d..957b1e5` is the original fresh-review input; head SHA matches the posted comment's `957b1e5`), run through the same read-only path. Nothing was posted to that PR.

```
decision = approve
gateRan  = True
issues   = 0
```

The same path previously produced `REVISE` + narration. Matches the `approve` the issue records for direct review.

**Regression tests** pin the posted Korean narration verbatim, English narration variants, a stream ending on `todo_list`, a sign-off after the verdict, a verdict-shaped mid-stream sentence, and the normal stream. Each fails against the old code.

**Gates**: 4196 tests · lint 0 · typecheck 0 · build OK. Pre-commit review gate ran 3 rounds (3 real defects found and fixed, including a `parseRecord` allow-list regression of my own), then escalated to an independent sub-agent review — **0 blocking**, one risk note which is addressed above.

## Out of scope

`freshReview` doesn't take `PR_PROCESS_LOCK_PATH`. It predates this diff, and adding the lease would change freshReview's documented parallel-safe design. Tracked as AGT-3916.

Closes AGT-3914.


unohee:
## 🔍 Fresh review of 4d868de (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The INT-3914 parser/history/gateRan fixes in 4d868de look solid and are well-tested, and the new quality/security harness wiring is largely fail-closed. Three material gaps remain: freshReview still mutates the shared checkout without PR_PROCESS_LOCK (tracked AGT-3916), pipeline guard new-file reads are still symlink-unsafe, and the newly registered Cursor chat provider cannot work because chatBackend ignores stdinFile and parses Codex NDJSON only.
  Suggestions:
    - Align reviewAudit.listSourceFiles (reviewAudit.ts:139-146) with qualityHarness git ls-files -z handling so area partitioning and harness inventory cannot diverge on unusual filenames.
    - Add composer-* routing in inferProviderFromModel (chatBackend.ts:110-121) so --model composer-2.5-fast without --adapter cursor selects the right backend.
    - Consider rejecting foreign model IDs in modelCompat for cursor (modelCompat.ts:50-52), mirroring claude/gpt guards, instead of passing codex/gpt slugs straight through to cursor-agent.
    - Evaluate whether Cursor read-only reviews should use --mode ask rather than --mode plan (cursor.ts:92-93) for parity with codex read-only sandbox enforcement on untrusted PR input.
  Recommended follow-ups:
    - [bug] Honor stdinFile and Cursor NDJSON parsing in chatBackend CLI spawn path (src/support/chatBackend.ts:335-507)
    - [bug] Fail closed when extractCursorResult returns empty in parseWorkerOutput (src/adapters/cursor.ts:121-123)
    - [test] Add chatBackend integration test for provider cursor (stdin wired, response parsed) (src/support/chatBackend.test.ts)

unohee:
## 🔍 Fresh review of 2194550 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  INT-3914 is addressed end-to-end: parseReviewerResult no longer fabricates REVISE from opening narration, Codex backward-scans agent_messages with a JSON-only bar for non-final messages, freshReview propagates gateRan correctly (including post-verdict comment failures), and PR review history is persisted/read from projectPath with kind:'pr' round-tripping. Tests cover the regression shapes and wiring overrides.

PR #413, opened 2026-08-27
chore(docker): modernize stale Docker packaging into a verified deployable image (AGT-3997)

unohee:
## TL;DR

The Docker packaging predated the OpenSwarm rename and had silently rotted — the image built a dashboard-less daemon with no CLI exec bit, health-checked an auth-gated endpoint, and persisted no state. This PR rewrites it into a locally build- and boot-verified deployable image, with a GHCR publish path and a packaging-scoped PR build check so it cannot rot silently again.

Closes AGT-3997.

## Changes

| File | Change |
|---|---|
| `Dockerfile` | Both stages `node:22-slim` (one libc — natives compiled once in the builder, copied to runtime; no runtime toolchain); `npm run build` (postbuild: CLI exec bit + `dist/web-static`); prod-pruned `node_modules`; git+gh+bubblewrap+dumb-init; non-root; `openswarm` on PATH; `/api/health` healthcheck; `safe.directory` for mounted repos |
| `docker-compose.yml` | `openswarm-state` volume (`~/.openswarm`), `/work` repos mount, `OPENSWARM_WEB_TOKEN` passthrough (dashboard binds container-internal without it — by design), config bind-mount footgun documented, ONNX/cairo-era leftovers removed |
| `.dockerignore` | Excludes `config.yaml`/`.env`/`dist`/state — secrets and stale host builds stay out of the build context |
| `.github/workflows/docker-publish.yml` | GHCR publish on release published + manual dispatch; buildx amd64+arm64; semver + pinned `latest=auto` tags |
| `.github/workflows/docker-build-check.yml` | PR build check scoped to packaging paths (build + CLI/assets smoke, never pushes) |
| `README.md` | "Run with Docker": quick start, mount matrix, token requirement, provider-CLI derivation, verify-sandbox flags |

## Verification (local, actual runs)

- `docker build` → success, 1.16GB (arm64; CI builds amd64+arm64)
- Container smoke: `openswarm --version` → 0.21.1; git/gh/bubblewrap present; non-root user
- Daemon boot on a minimal config → `/api/health` 200 (token-less), dashboard `/` 200, `/static/js/api.mjs` 200
- Auth posture: `/api/coordination` 403 without token, 200 with `X-OpenSwarm-Token`
- Docker `HEALTHCHECK` → `healthy`; state dir written by the native sqlite addon (glibc prebuild works)
- `docker compose config` clean; workflow YAML parse clean
[...]

PR #412, opened 2026-08-27
feat(coordination): autonomous agent collaboration, human-in-the-loop, MCP orchestration (AGT-3993)

unohee:
## TL;DR

Wires OpenSwarm's isolated worktree workers into one coordination plane: a durable agent board with deterministic Warhammer-40k call signs, Discord human-in-the-loop for blocking decisions, typed-failure adapter routing (codex-responses → cc-router/cursor), leased read-only periodic review agents, and a least-privilege MCP-connected orchestrator — all supervisable from the web dashboard.

Closes AGT-3993.

## What's in here

| Area | Change |
|---|---|
| Coordination board | `src/coordination/coordinationStore.ts` — file-locked, fingerprint-deduped, secret-redacted event store; Linear issue mirror via `coordinationBoardIssueId`; `coordination_read`/`coordination_publish` agent tools |
| Call signs | `src/coordination/agentNames.ts` — deterministic per repository/task/role, so a name survives restarts and doubles as the mailbox address; worker ≠ reviewer on one task |
| Human-in-the-loop | `ask_human` → Discord page (`!answer <correlation-id>`); the agentic loop hard-stops on a blocking question and the pipeline stops instead of retrying into it (`blockedOnOperator`) |
| Adapter routing | `adapterRouting` gated on typed reasons (quota/infra/capability); semantic failures never provider-hop; new `cc-router` and `cursor` adapters incl. chat-backend stdin + stream-json parsing |
| Periodic reviews | `periodicReviews` profiles (permissions/hygiene/security/review), read-only, leased per repository+profile — never overlap |
| Orchestrator | Per-role MCP grants (github/linear/cloudflare), denied tools audited to the board; scratch cwd **and** `shellTools: false` (bash is not path-checked); `spawnCli` fails closed when a delegated CLI can't honor that |
| Instruction capsule | Claude Code rule hierarchy snapshotted once per run, identical for every role, digest shown on the dashboard |
| Supervision | Dashboard `AGENT COORDINATION` panel, `GET /api/coordination`, SSE `coordination:event`; UI freeze cancelled in `docs/PLATFORM_ROADMAP.md` |

## Verification

- Full suite: **310 files / 4,201 tests passed** (plus 1 skipped), consecutive clean runs
- `tsc --noEmit`, `oxlint` (0/0), `npm run build` (+x on `dist/cli.js`) clean
[...]

PR #414, opened 2026-08-27
test(adapters): cover the cc-router and cursor adapters (AGT-3993 follow-up)

unohee:
## TL;DR

The AGT-3993 merge landed `cc-router` with **zero** function coverage and `cursor` at 3/11, dropping global function coverage to 89.67% — under the 90% gate. Main's CI is red and every open PR (including #413) is blocked on it. This adds the missing adapter tests; coverage recovers to **90.45%**.

The PR-side Tests check on #412 was misread as green because the job registered late while the merge decision was taken — the gate itself was correct. Process note recorded; the fix here is the tests the new code should have shipped with.

## What's covered

- **cc-router availability contract**: `ok` status AND `openAIResponses` capability required — a reachable router that can't serve the Responses API is an unusable primary, and capability routing depends on this probe being strict
- **Identity**: bearer token only when `CC_ROUTER_TOKEN` is set, never a ChatGPT account id
- **Egress guard**: local `/v1/responses` admitted, remote base URL rejected
- **Model discovery**: id extraction, banner filtering (cursor), env/hardcoded fallbacks, non-OK → `[]`
- **Cursor streaming**: partial-line buffering across chunks (no lost or duplicated events), worker/reviewer result parsing with executed-command merge

## Verification

- `npm run test:coverage`: functions **90.45%** (2357/2606), 311 files / 4,217 tests green
- lint 0/0, pre-commit typecheck clean
- Tests-only diff — per the commit-gate policy the tier-1 review is not required for it

### DIFF
--- a/src/adapters/ccRouter.test.ts
+++ b/src/adapters/ccRouter.test.ts
@@ -1,0 +1,124 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const execFileMock = vi.hoisted(() => vi.fn());
+vi.mock('node:child_process', () => ({ execFile: execFileMock }));
+
+import { CcRouterAdapter } from './ccRouter.js';
+
+// The adapter awaits promisify(execFile); the generic promisify wrapper
+// resolves with the callback's first success value, so the mock hands back the
+// { stdout } object the real promisified execFile would.
+type ExecCb = (err: Error | null, result?: { stdout: string; stderr: string }) => void;
+function statusReply(payload: unknown): void {
+  execFileMock.mockImplementation((_c: string, _a: string[], _o: unknown, cb: ExecCb) =>
+    cb(null, { stdout: JSON.stringify(payload), stderr: '' }));
+}
+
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 5757ffa (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Test-only diff adds focused unit coverage for CcRouterAdapter and CursorCliAdapter that matches current production behavior: capability-gated availability, model discovery/fallback, loopback egress rejection, stream buffering, and result parsing. The execFile mock shape ({ stdout, stderr } as a single callback success value) is correct under vi.mock because promisify runs on the mock, not Node’s custom execFile promisifier. No material correctness or security defects found in the changed files.
  Suggestions:
    - In ccRouter.test.ts and cursor.test.ts, assert execFile invocation args (e.g. cc-router status --json, cursor-agent --list-models) so CLI contract regressions are caught, not only return-shape behavior.
    - Add a ccRouter base-URL case without a /v1 suffix (default http://127.0.0.1:3456) alongside the existing /v1/ normalization test.
    - Use vi.restoreAllMocks() in cursor.test.ts afterEach (as ccRouter.test.ts already does) for consistent spy cleanup if fetch or other globals are spied later.

PR #417, opened 2026-08-27
feat(dashboard): readable event feed, agent conversations, operator replies (AGT-4006)

unohee:
## Why

The orchestration feed rendered `event.summary` and nothing else. A real row on the live vela deployment read:

```
Claude Code rules c480ceccd832 (0 sources)
```

No time, no task, no speaker, and `detail`/`metadata` were dropped entirely. `taskId` is an issue UUID, so nothing on screen said which issue an event belonged to. There was also no way to read a request and its answer as one exchange, and no way to speak into one.

## What changed

**Legible rows.** Events carry a new `taskLabel`, stamped by the publishers that know the issue identifier and deliberately excluded from the dedup fingerprint — a display name must not split content dedup. Each row now states the time, the task, and who spoke to whom as a sentence, with the long form and metadata shown instead of discarded.

**Conversations.** Clicking a row opens the exchange it belongs to as a transcript and rings its speaker in the graph; the addressee gets a quieter dashed ring, so one message reads against the hierarchy.

**Operator replies.** The transcript carries a composer. Replying to a blocking question routes through `answerHumanQuestion` and unblocks the agent that asked; any other reply is published addressed to the last agent speaker, which `coordination_read` then delivers. This is a real channel into a running task, not a comment box.

**Durable trace.** The board is a 2000-event ring buffer, so it forgets. Every published event is now archived append-only in the SQLite database that already holds the run ledger, so traces and runs stay joinable in one engine and survive container recreation on the mounted state volume. Agents read it back with a new `coordination_history` tool; the dashboard uses `GET /api/coordination/history`. Archiving is best-effort by construction — a trace failure never fails the publish that produced the event.

## Security

Mutations inherit web.ts's existing gate, which rejects unauthorized `/api/` writes *before* route delegation. Verified at the call site rather than duplicated. All event-fed strings stay `escapeHtml`-wrapped, with a test asserting script/img/iframe payloads never materialize.

## Verification

- `tsc --noEmit`, `oxlint src/`: clean
- `vitest run`: 316 files, 4266 passed / 1 skipped
- New: 6 trace tests (including reading back an event the ring buffer evicted), 16 conversation-model tests, 4 interjection route tests (asserting `consume()` actually hands the message over), 6 view tests (feed content, XSS escaping, speaker highlight, transcript, composer payload)
- `npm run build` + assets confirmed in `dist/web-static/`
- Commit gate: tier-1 `openswarm review` → `Decision: APPROVE`

### DIFF
--- a/src/agents/pairPipeline.ts
+++ b/src/agents/pairPipeline.ts
@@ -85,7 +85,14 @@ import { stageTimeoutMs } from './stageTimeouts.js';
 function coordinationContextFor(context: PipelineContext, role: AgentRole) {
   const taskId = taskEventKey(context.task);
   const callSign = assignCallSign({ repository: context.projectPath, executionId: taskId, role });
-  return { repository: context.projectPath, taskId, actor: callSign.address, actorName: callSign.name, actorRole: role };
+  return {
+    repository: context.projectPath,
+    taskId,
+    taskLabel: context.task.issueIdentifier,
+    actor: callSign.address,
+    actorName: callSign.name,
+    actorRole: role,
+  };
 }
 
 export class PairPipeline extends EventEmitter {
--- a/src/automation/automationDbPath.ts
+++ b/src/automation/automationDbPath.ts
@@ -1,0 +1,17 @@
+// ============================================
+// OpenSwarm - Automation database path
+// ============================================
+//
+// A leaf module with no native dependencies, so consumers can learn where the
+// automation database lives without importing `runLedger` — which pulls in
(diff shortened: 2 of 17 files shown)


unohee:
## 🔍 Fresh review of a168941 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Reviewed the committed coordination trace, history/message API, task-label propagation, and orchestration UI wiring. No concrete correctness or production-wiring defect was identified from the changed code.

PR #418, opened 2026-08-27
fix(dashboard): stop long metadata values from scrolling the feed sideways (AGT-4006)

unohee:
## Why

Reported from the live vela dashboard: the event feed had a horizontal scrollbar under it.

Cause: metadata chips rendered their value raw, and a content digest is a 64-character unbreakable token — one `instruction-snapshot` row was wider than the sidebar and dragged the whole feed sideways.

## What changed

- Chips truncate with an ellipsis and keep the full `key: value` in their `title`, so the digest is still readable on hover rather than lost.
- Feed and transcript scroll vertically only (`overflow-x: hidden`), and their flex children are pinned to `min-width: 0` — without that, `min-width: auto` lets one long word widen the panel and the ellipsis never applies.
- Long words in the summary/detail lines break (`overflow-wrap: anywhere`) instead of pushing.
- The remaining vertical scrollbars are thin and panel-coloured; the default chrome read as a defect against the dark surface.

## Verification

- `vitest run tests/web/orchestrationView.test.ts`: 12 passed, including a new case asserting the tooltip is HTML-escaped (metadata is event-fed, so it is untrusted).
- Pre-commit gate: oxlint + typecheck clean.

### DIFF
--- a/tests/web/orchestrationView.test.ts
+++ b/tests/web/orchestrationView.test.ts
@@ -179,6 +179,9 @@ describe('feed legibility', () => {
     const row = doc.querySelector('#feed .ev')!;
     expect(row.querySelector('.ev-line')!.textContent).toBe('OpenSwarm daemon (daemon) loaded the rule set');
     expect(row.querySelector('.ev-meta')!.textContent).toContain('digest');
+    // The chip truncates visually, so the full value has to survive somewhere
+    // the operator can still reach — otherwise a digest is unreadable.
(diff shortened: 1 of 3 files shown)


unohee:
## 🔍 Fresh review of 1b3d3c8 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Metadata tooltip content is HTML-escaped before interpolation, preserving the full chip value without introducing attribute injection. The CSS constrains long feed content to the aside and tests cover both tooltip preservation and hostile key/value input.

PR #419, opened 2026-08-27
fix(security-audit): report why the audit failed instead of just "failed" (AGT-4010)

unohee:
## Why

A swarm run on the live vela deployment stopped with exactly this:

```
Pipeline infra error (not counted toward STUCK): security-audit-infra: failed
[Runner] PR not created for AGT-4008: Pipeline failed (infra_error)
```

That word is the entire diagnosis. CodeQL itself was healthy — I reproduced `codeql database create` (660 files extracted) and `codeql database analyze` (105 queries, SARIF written) inside the same container by hand, both clean — so the operator is told the audit failed while every visible signal says it should not have.

## Cause

The cause is recorded and then dropped twice:

1. `runSecurityAudit`'s catch-all writes the real error into `findings[0].message` but leaves `detail` unset.
2. `SecurityAuditInfrastructureError` builds its message from `result.detail`, then falls through to `result.status` — skipping the findings entirely. With `detail` unset, `status` is `'failed'`, and that is what reaches the log.

Every other failure path in `runSecurityAudit` sets `detail`, which is why only unexpected exceptions surface this way — exactly the failures where a cause matters most.

## What changed

- The catch-all sets `detail` to the same cause it puts in the finding.
- The gate consults the first error-level finding before settling for `status`.

## Verification

- 4 new tests covering the precedence chain: explicit detail wins, an error finding is used when detail is absent, non-error findings are ignored, and a missing result still yields a message.
- `vitest run src/verify/securityAudit`: 15 passed (unchanged).
- Pre-commit gate: oxlint + typecheck clean.

Note: this makes the underlying AGT-4008 failure diagnosable; it does not claim to fix it. The next run will name the actual cause.

### DIFF
--- a/src/agents/securityAuditGate.diagnostics.test.ts
+++ b/src/agents/securityAuditGate.diagnostics.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it } from 'vitest';
+import { SecurityAuditInfrastructureError } from './securityAuditGate.js';
+import type { SecurityAuditResult } from '../verify/securityAudit.js';
+
+function result(over: Partial<SecurityAuditResult> = {}): SecurityAuditResult {
+  return {
+    status: 'failed',
+    codeqlLanguages: ['javascript-typescript'],
+    skippedCodeqlLanguages: [],
+    findings: [],
+    ...over,
+  } as SecurityAuditResult;
+}
+
+describe('SecurityAuditInfrastructureError', () => {
+  it('prefers an explicit detail', () => {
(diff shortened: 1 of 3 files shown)


unohee:
## 🔍 Fresh review of 0b2e62d (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  The gate now exposes actionable audit failure context: it prefers result.detail, then an error finding, before status (src/agents/securityAuditGate.ts:16-17). Unexpected runSecurityAudit exceptions now populate both the runtime finding and detail (src/verify/securityAudit.ts:333-336); focused diagnostics tests cover precedence and fallbacks.

PR #420, opened 2026-08-27
fix(docker): install python3 so the security audit can analyse Python (AGT-4011)

unohee:
## Why

Every swarm task on the container deployment parked at the security gate:

```
Pipeline infra error (not counted toward STUCK): security-audit-infra: failed
[Runner] PR not created for AGT-4008: Pipeline failed (infra_error)
```

The gate fails closed, so this stopped the autonomous loop before the worker stage ever ran. No agent reached the board, which is why the orchestration view had nothing real to show.

## Cause

CodeQL was healthy — inside the same container I ran `codeql database create` (660 files extracted) and `codeql database analyze` (105 queries, SARIF written) by hand, both clean.

Reproducing `runSecurityAudit` itself and reading its findings named the real cause:

```
error | openswarm/security-codeql-python-database | CodeQL could not create its no-build security database.
```

The audit analyses every language present in the repository. This repository tracks one `.py` file (`src/task_state_model.py`), so CodeQL runs its Python extractor — and that extractor shells out to an interpreter which a `node:22-slim` runtime does not ship. `which python3` in the container returned nothing.

The builder stage already installs `python3` (for native module builds); the runtime stage did not.

## Verification

Installed `python3` into the running container and re-ran the identical audit:

| | before | after |
|---|---|---|
| status | `failed` | `findings` |
| infrastructure findings | `openswarm/security-codeql-python-database` | none |

`findings` is the healthy outcome — the audit completed and reported real code findings rather than an infrastructure failure.

## Related

PR #419 (merged) is what made this diagnosable: the cause was recorded in the findings but discarded before reaching the log.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,9 +43,14 @@ FROM node:22-slim AS production
 # git + gh: workers commit and open PRs. bubblewrap: the verify sandbox —
 # fail-closed under Docker's default seccomp profile; see README for the flags
 # that enable it. curl: healthcheck. dumb-init: PID-1 signal handling.
+# python3 is a runtime dependency, not a build one: the security-audit gate runs
+# CodeQL over every language present in the analysed repository, and CodeQL's
+# Python extractor shells out to an interpreter. Without it a single tracked
+# .py file makes `codeql database create --build-mode=none` fail, which the
+# pipeline reports as an infra error and parks the task.
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-        dumb-init ca-certificates curl git bubblewrap gnupg && \
+        dumb-init ca-certificates curl git bubblewrap gnupg python3 && \
     curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
         -o /usr/share/keyrings/githubcli-archive-keyring.gpg && \
     chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg && \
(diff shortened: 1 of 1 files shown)


unohee:
## 🔍 Fresh review of 0f5c3ea (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  The production image now installs the Python interpreter required by CodeQL's Python extractor. This matches the audit's per-language database creation path and prevents Python-containing target repositories from failing the security-audit gate due to a missing runtime interpreter.

unohee:
## 🔍 Fresh review of 0f5c3ea (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Adding `python3` is insufficient to make the Docker security-audit gate work: the production image still does not install or copy a `codeql` executable. The audit therefore returns `unavailable` before Python extraction is reached.

unohee:
Addressed in `039929e`.

The finding is correct: the image does not ship the CodeQL CLI, so a bare deployment reports `unavailable` before Python extraction is reached. That omission is deliberate — the bundle is roughly a gigabyte and would be carried by every pull — but it was undocumented, which is the actual defect. The running deployment already provides it by mount (`/opt/codeql:ro` plus a PATH entry); that contract is now recorded in the Dockerfile next to the CodeQL config the image does carry.

The python3 change stands on its own: with CodeQL present and python3 absent, the audit fails at Python database creation, which is exactly what parked every task on the live deployment. Measured in the running container before and after installing python3:

| | before | after |
|---|---|---|
| status | `failed` | `findings` |
| infrastructure findings | `openswarm/security-codeql-python-database` | none |

PR #421, opened 2026-08-27
fix(linear): report a failed issue lookup instead of calling it "not found" (AGT-4012)

unohee:
## Why

The daemon's Linear OAuth token expired on the container deployment. Every dispatch then returned:

```json
{"issueId": "AGT-4008", "status": "skipped", "reason": "not found"}
```

The issue existed the whole time. Finding that out took reading the daemon's stderr, where the real error was sitting:

```
[Linear] getIssue error: AGT-4008 Error: Authentication required, not authenticated
```

`getIssue` catches every API error and returns `null`, and `null` is indistinguishable from a missing issue at the call site. So an expired credential was reported as a missing issue — pointing the operator at Linear's contents rather than at the daemon's authentication. For an autonomous loop this is the difference between "someone deleted my ticket" and "re-authenticate".

## What changed

- `lookupIssue` returns `{ ok: true, issue }` or `{ ok: false, error }`, separating "no such issue" from "the lookup could not be performed".
- `getIssue` delegates to it and keeps its `null` contract, so callers that only need presence are untouched.
- `dispatchWork` uses `lookupIssue` and reports `Linear lookup failed: <cause>`.

## Verification

- 2 new tests: a lookup failure names the credential error rather than the issue; a genuinely missing issue still reports `not found`.
- `vitest run src/automation/workRunner`: 25 passed. The existing dispatch suite mocked `getIssue`, so its mock now provides `lookupIssue` delegating to it — the assertions on `getIssue` are unchanged and still exercise the production path.
- `vitest run src/linear/`: 20 passed.
- oxlint + typecheck clean.

### DIFF
--- a/src/automation/workRunner.lookup.test.ts
+++ b/src/automation/workRunner.lookup.test.ts
@@ -1,0 +1,65 @@
+import { afterAll, beforeEach, describe, expect, it, vi } from 'vitest';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+// dispatchWork checks the project path exists before any lookup, so the test
+// needs a real directory rather than a placeholder string.
+const repo = mkdtempSync(join(tmpdir(), 'work-dispatch-'));
+afterAll(() => rmSync(repo, { recursive: true, force: true }));
+
+const linear = vi.hoisted(() => ({
+  isLinearInitialized: vi.fn(() => true),
+  lookupIssue: vi.fn(),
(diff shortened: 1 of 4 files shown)


unohee:
## 🔍 Fresh review of e855058 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Dispatch now distinguishes a failed Linear lookup from a confirmed absent issue, while `getIssue` preserves its existing null-on-failure contract. Production wiring is present in `dispatchWork`, and the added tests cover lookup failure and confirmed absence.

PR #425, opened 2026-08-28
fix(adapters): prove process-group ownership before any group signal (AGT-4011)

unohee:
## TL;DR

`signalCliProcessTree` ran `process.kill(-proc.pid, signal)` unconditionally. Unit tests hand fabricated pids (`1`, `321`) to close handlers, so the resulting `kill(-1, SIGKILL)` signalled **every process the user may signal**: it wiped the operator's macOS login session twice, severed the SSH connection running the suite on a remote host, and killed the GitHub runner mid-suite — the cause of `Tests=CANCELLED` on #421–#424.

## Design: three-layer ownership proof

1. **Live leader path** — `kill(-pgid)` fires only after the OS confirms *at kill time* that the pid leads its own group (`pgid == pid`) **and** is our direct child (`ppid == process.pid`), via `/proc/<pid>/stat` on Linux and `ps -o ppid=,pgid=` elsewhere. Pid 1, our own pid, and our own group are refused before any lookup.
2. **Track-time record** — `trackCliProcessTree` proves the same ownership right after spawn (the only moment it is provable: a detached child is a live leader whose parent is us) and stamps the leader's death with `once('exit')`. The record lives in a WeakMap keyed by the handle, so callers that untrack before terminating (codexUserMcp) cannot drop it.
3. **Post-mortem sweep** — once the leader is dead, `kill(-pgid)` is **never** used (a recycled group id cannot be re-verified). Orphaned members are enumerated with their ages (`/proc` starttime on Linux, `ps -A -o pid=,pgid=,etime=` on macOS) and signalled one pid at a time, only when their age proves they started before the leader died — every member of a recycled group is younger than that. Clock fuzz matches measurement granularity (250ms Linux / 1500ms macOS). Enumeration or verification failure means **leak, never misfire**.

Residual risk: a recycled group id inside the fuzz band would require the kernel to exhaust its entire pid space within it (≥131k forks/s on Linux at the 32768 floor). What remains is the same per-pid check-then-signal window the live path — and every kill-by-pid tool — already carries.

## Also in this PR

- `codexUserMcp.test.ts`: on macOS a freshly written script takes 200–400ms to first-exec (measured; Gatekeeper scan), longer than the old 100ms timeout — the fixture was killed before it started. A warm-up exec + 500ms timeout keeps the assertion (a TERM-ignoring child must die) intact.
- `processTree.groupSafety.test.ts` (new): 10 tests pinning the ownership proof — fabricated pids, non-leaders, dead pids, recycled groups, per-pid orphan sweep.
- `base.spawn.test.ts`: injection contract updated to ownership objects; one test renamed to match what it now asserts.

## Verification

- 58/58 across the 4 affected test files on macOS, including two real-process e2e paths (post-mortem sweep via real `ps`, live-path group kill of a TERM-ignoring child); `tsc --noEmit` clean.
- `/proc` field indices (ppid/pgrp/starttime) and elapsed-time math verified empirically on Linux (vela).
- Review gate: tier-1 `openswarm review` ×3 (REVISE, converging on the pid-recycle residual) → tier-2 independent subagent: **APPROVE**, with the residual risk explicitly quantified and accepted; its one Low finding (misleading test name) is fixed in this PR.

Closes AGT-4011.

### DIFF
--- a/src/adapters/base.spawn.test.ts
+++ b/src/adapters/base.spawn.test.ts
@@ -23,7 +23,9 @@ describe('CLI process tree termination', () => {
     const processKill = vi.spyOn(process, 'kill').mockImplementation(() => true);
     const directKill = vi.fn(() => true);
 
-    terminateCliProcessTree({ pid: 7654, kill: directKill } as never, 'linux');
+    // The lookup answers "7654 leads its own group and is our child": only
+    // then may the whole group be signalled.
+    terminateCliProcessTree({ pid: 7654, kill: directKill } as never, 'linux', () => ({ pgid: 7654, ppid: process.pid }));
 
     expect(processKill).toHaveBeenCalledWith(-7654, 'SIGKILL');
     expect(directKill).not.toHaveBeenCalled();
@@ -35,7 +37,7 @@ describe('CLI process tree termination', () => {
     });
     const directKill = vi.fn(() => true);
 
-    terminateCliProcessTree({ pid: 7655, kill: directKill } as never, 'linux');
+    terminateCliProcessTree({ pid: 7655, kill: directKill } as never, 'linux', () => ({ pgid: 7655, ppid: process.pid }));
 
     expect(directKill).toHaveBeenCalledWith('SIGKILL');
   });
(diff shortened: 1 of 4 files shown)


unohee:
## 🔍 Fresh review of 4c0e182 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  The POSIX group signal now requires a live direct-child group leader (src/adapters/processTree.ts:537-545), and post-exit cleanup uses per-PID age-gated sweeping rather than an unsafe negative-PGID signal (src/adapters/processTree.ts:550-564). Production spawn paths track processes before lifecycle cleanup, while the WeakMap retains verified ownership through untracking; targeted tests cover fabricated, self, unrelated, live, and exited PID cases.

PR #428, opened 2026-08-28
fix(task-state): free a lock whose owner cannot be probed across a restart (AGT-4023)

unohee:
## The incident (measured, on vela)

A container recreate killed the daemon mid-operation and left a well-formed lock at `~/.openswarm/task-state.json.lock` — `{"pid":7,...}` — on the bind-mounted state volume. A container assigns the daemon **the same pid every start**, so `isProcessAlive(7)` probed the *new* daemon and answered "held". Nothing could ever free it.

Measured: **15 consecutive heartbeats** failed with `Timed out waiting for task state lock`, the daemon fetched **zero** tasks for **75 minutes**, `runningTasks=0`, and a manual `rm` was the only recovery. Confirmed after removal: next heartbeat fetched 755 tasks and started work.

## Fix

Identity cannot settle this. A pid-namespace boundary makes "my pid from an earlier generation" and "a live process elsewhere sharing the mounted state file" indistinguishable — an instance-id scheme was tried and withdrawn for exactly that reason. Age is the one signal true across namespaces and generations: a lock older than `LOCK_ABANDON_MS` (10 min) is released regardless of its owner field. The two pre-existing rules are unchanged.

## The tradeoff, stated rather than claimed away

The code comment now says what this actually costs instead of asserting it is free:

- Eviction requires a process **frozen for ten minutes** (SIGSTOP / docker pause / uninterruptible I/O) or a wall-clock jump. The guarded work is synchronous — read, parse, write, fsync, rename, measured at tens of ms — and an out-of-space write fails fast rather than blocking.
- If an evicted holder resumes it overwrites with its own snapshot: **a lost update, never a torn file** (`persistStore` renames atomically) and **never a cascade** (the exit unlink is token-guarded).
- The **run ledger owns leases and remote effects**, not this projection, so a rollback here cannot double-execute anything; the next Linear sync re-derives it.

Against a *certain, measured* 75-minute outage needing a human.

## Review

The tier-1 gate raised mutual exclusion four times. An independent tier-2 review adjudicated it with evidence — ENOSPC returns an error rather than blocking (so the cited trigger releases in ms), `upsertTaskState` is the sole `withStoreLock` caller and is fully synchronous, and `runLedger.importRun` is insert-only for leases — and **approved the design unchanged**, asking only for three mechanical fixes, all applied:

- the comment no longer overclaims (it now names the frozen-process and non-monotonic-clock exposure);
- a test pins the threshold **from below** so it cannot quietly collapse into the 30s malformed clock — mutation-verified: setting 600s → 60s fails the suite;
- a duplicate 5s test was dropped (suite 10.9s → 5.9s).

Follow-up filed separately for the kernel-owned SQLite mutex (`serviceInstanceLock.ts` pattern), which removes thresholds entirely but needs a CLI/daemon migration story.

## Test plan

`npx tsc --noEmit` clean; `LC_ALL=C npx vitest run src/taskState` 17/17 including the two-process serialization test; oxlint clean.

## Linear

Closes AGT-4023

### DIFF
--- a/src/taskState/store.test.ts
+++ b/src/taskState/store.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
-import { existsSync, mkdtempSync, readFileSync, rmSync, unlinkSync, writeFileSync } from 'node:fs';
+import { existsSync, mkdtempSync, readFileSync, rmSync, unlinkSync, utimesSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
 import { spawn } from 'node:child_process';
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 8dfcba4 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The new expiry path reclaims locks held by a potentially live process, but its unconditional deletion is not safe against concurrent lock turnover. The tests cover expiry timing only and do not exercise this interleaving.

unohee:
## Re: the repeated mutual-exclusion REVISE — shipping with the tradeoff documented

Five tier-1 rounds have raised one theme: an advisory dot-lock cannot be made perfectly safe. **That is true**, and it is why AGT-4024 is filed for the kernel-owned SQLite mutex (`serviceInstanceLock.ts` already uses `BEGIN IMMEDIATE` for exactly this reason). It is not a reason to keep the current behaviour, which is a *certain* outage.

What each round produced:

| Round | Finding | Outcome |
|---|---|---|
| 1–4 | "expiry can evict a live writer" | Adjudicated by an independent tier-2 review; design approved unchanged, three mechanical fixes applied |
| PR review | reclaim→unlink turnover race | **Accepted and narrowed** — `3e7fbcb` re-reads mtime + token immediately before the unlink |
| 5 | the narrowed window is still not zero | True, inherent to POSIX; documented at the decision site and tracked in AGT-4024 |

The tier-2 adjudication of rounds 1–4, with evidence:

- The cited trigger (**fsync on a full disk**) does not stall — ENOSPC *returns an error*; `persistStore` throws, the `finally` unlinks, and the holder releases in milliseconds. The named disaster is the case the code already handles fastest.
- `upsertTaskState` is the **sole** `withStoreLock` caller and is entirely synchronous (read, parse, write, fsync, rename — measured 15–33 ms). Reaching 10 minutes requires a SIGSTOP/pause/D-state freeze, by which point that single-threaded process has served zero heartbeats.
- Worst case if such a corpse resumes: a **lost update to a projection** — never a torn file (atomic rename), never a cascade (token-guarded exit unlink). `runLedger.importRun` is insert-only and refuses lease/effect states from legacy files, so a rollback here **cannot** double-execute a task; the next Linear sync re-derives it.

The race the PR review found is **pre-existing** — both prior reclamation rules had it — and is now strictly narrower than before this PR, not wider.

Measured status quo being fixed: 15 consecutive dead heartbeats, 755 tasks unfetched, 75 minutes of zero throughput, human `rm` the only exit. Merging on green CI.

PR #429, opened 2026-08-28
fix(web): stop losing the operator's message without saying so (AGT-4026)

unohee:
## The report

Operator, with **nine agents parked on `human-question`**: the orchestration view shows "awaiting your answer", typing a reply does nothing — no message, no error.

## Two defects on the same line, in both composers

```js
input.value = '';        // cleared before the send is known to have worked
send(text);              // not awaited
try { await fetcher('/api/coordination/message', {...}); }
catch { /* the next poll shows whether it landed */ }
```

1. **`fetch` does not reject on 4xx/5xx.** The daemon's own refusals — `400 Cannot address the message`, `409 Question is already completed` — arrived in the success path. Both are returned by the live endpoint today.
2. **The box was cleared before the outcome was known**, so a refused reply was also text the operator had to retype from memory.

## Measured, and why the in-flight state matters

| condition | POST /api/coordination/message |
| --- | --- |
| no tasks running | 9 ms |
| 2 tasks running | 8.3 s |
| 2 tasks running (earlier) | >20 s, client aborted |

Publishing queues behind the agents' own board traffic (filed as AGT-4027). So even a reply that lands looks dead for seconds without a pending state.

## What changed

Both composers now check `response.ok`, surface the daemon's `error` text, keep the operator's words until the send is accepted, and show "Sending…" while it is in flight.

Two things the review rounds surfaced, both fixed and pinned by tests:

- **The pending state lives in the view, not the form.** Agent events arrive constantly; a redraw mid-send replaces the composer, which would otherwise re-enable the box and drop the text.
- **It is kept per exchange.** The operator can move to another thread while a send is running and start a second there; one completing must not land its verdict on the other's composer.

## Test plan

`npx tsc --noEmit` clean, oxlint clean, `tests/web` **165/165** — including a refused send (keeps text + shows reason), an unreachable daemon, a gated in-flight send (locked box, text intact), and a cross-exchange leak check.

Gate: `openswarm review` 4 rounds → APPROVE (rounds 1–3 found the redraw race, the thread-scoping, and the concurrent-send overwrite).

## Linear

Closes AGT-4026

### DIFF
--- a/tests/web/chatView.test.ts
+++ b/tests/web/chatView.test.ts
@@ -40,10 +40,10 @@ function boardEvent(over: Record<string, unknown> = {}) {
   };
 }
 
-function fetchWith(history: unknown[], board: unknown[] = []) {
+function fetchWith(history: unknown[], board: unknown[] = [], messageResponse?: unknown) {
   return vi.fn(async (url: string) => {
     if (url.startsWith('/api/coordination/message')) {
-      return { ok: true, json: async () => ({ delivered: true }) };
+      return messageResponse ?? { ok: true, json: async () => ({ delivered: true }) };
     }
     if (url.startsWith('/api/coordination/history')) {
       return { ok: true, json: async () => ({ events: history, traceSize: history.length }) };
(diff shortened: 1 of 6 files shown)


unohee:
## 🔍 Fresh review of 82ec86d (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  The views now retain text and expose server/network failures for non-OK POST outcomes, while preserving pending state across orchestration redraws via exchange-keyed composer state. Changed tests exercise accepted, refused, unreachable, and pending-send UI outcomes; no material defect found in the current diff.

PR #430, opened 2026-08-28
fix(web): send the operator's orchestration reply as a POST (AGT-4029)

unohee:
## The bug

```js
const fetcher = fetchImpl ?? ((url) => fetch(url));   // init dropped
```

Every send passes `{ method: 'POST', headers, body }` as the second argument. The default fetcher threw it away, so in a browser the operator's reply went out as a **GET** — and that path has no GET route:

```
GET  /api/coordination/message → 404 Not Found     (measured, live)
POST /api/coordination/message → 400 / 409 / 202   (measured, live)
```

`chatView.mjs` already forwarded it, which is exactly why the operator hit this only in the orchestration view.

This is the root cause of the original report — *typing a reply does nothing*. AGT-4026 stopped the composer swallowing failures, which turned the silence into a visible `The daemon refused the message (404)` and pointed straight here.

## Why no test caught it

Every existing test injects `fetchImpl`, so the fetcher a browser uses was never exercised. The tests assert the POST init the view *builds* — what was untested is whether anything forwards it.

The new test drives the view with **no** `fetchImpl`, stubs `globalThis.fetch`, and asserts the method and body arrive. Mutation-verified: restoring the old one-liner fails it.

## Test plan

`npx tsc --noEmit` clean, `tests/web` **166/166**, `openswarm review` APPROVE.

## Linear

Closes AGT-4029

### DIFF
--- a/tests/web/orchestrationView.test.ts
+++ b/tests/web/orchestrationView.test.ts
@@ -429,3 +429,41 @@ describe('clicking a feed row', () => {
     view.stop();
   });
 });
+
+describe('the default fetcher (AGT-4029)', () => {
+  // Every other test injects fetchImpl, so the fetcher a real browser uses was
+  // never exercised — and it dropped the request init, turning each POST into a
+  // GET the daemon answered 404. Drive the view with no fetchImpl so the
+  // default path is the one under test.
+  it('forwards method and body, so a send is a POST and not a GET', async () => {
+    const doc = shell();
+    const events = [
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 5eb4f37 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  The default fetcher now forwards the request init at web/static/js/orchestrationView.mjs:370, preserving the POST method, headers, and JSON body passed by publish() at lines 403-413. The new test exercises the non-injected browser fetch path and verifies the actual message request is POST with the composed body.

PR #431, opened 2026-08-28
feat(hitl): let the operator actually answer a parked agent (AGT-4030)

unohee:
## Why

Ten agents were sitting on `human-question` with *awaiting your answer* next to them, and **neither surface could answer one**. The composer addressed the newest exchange, so the daemon — which only converts a message into a `human-answer` when its correlation id names a pending question — filed the reply as an ordinary note. Nothing unparked.

This is the last of three defects that together left human-in-the-loop with no working entry point: AGT-4029 (the orchestration composer sent GETs), AGT-4026 (failures were swallowed silently), and this one (replies went to the wrong exchange).

## What changed

**The decision moved to the server.** The dashboard reasons only over the events it has loaded — a window on a ring buffer — so a question older than that window is invisible to it. The retained board is on the server, so `findOpenQuestionFor` lives there and the route prefers it when the client's named exchange is not a pending question.

**The client still resolves it too**, so the composer can say *which* question the next message will answer instead of leaving the operator guessing.

**Both lookups are scoped to a repository/task, and refuse when given neither.** Agents name themselves, so two on different tasks can both answer to `sable`, and the chat room shows every task at once — unscoped, an operator could unpark the wrong agent with an answer meant for someone else.

**Newest open question wins.** A run stops at the question it asked, so the latest unsettled one is what the agent is parked on and what the operator is reading.

## Verified

Answer replay does *not* depend on the mailbox address — `humanQuestionCorrelation` hashes (repository, taskId, question), so the next run of a task finds the recorded answer by correlation id, and the `waiting_on_operator` handler already re-admits the task on a backoff timer. That is why routing the reply onto the question's exchange is sufficient to unpark it.

## Test plan

`npx tsc --noEmit` clean, lint clean, **231/231** across `src/coordination` and `tests/web` — including answering past a stale client hint, refusing a same-named agent on another task, picking the newest of several open questions, and leaving an ordinary note alone.

`openswarm review`: 7 rounds → APPROVE (rounds found the settled-newest miss, the ring-buffer window, cross-task collision on both sides, and a comment that contradicted the code).

## Linear

Closes AGT-4030

### DIFF
--- a/src/coordination/coordinationRoutes.test.ts
+++ b/src/coordination/coordinationRoutes.test.ts
@@ -93,6 +93,105 @@ describe('POST /api/coordination/message', () => {
     expect(delivered).toMatchObject({ recipient: 'worker-a', detail: 'Use the existing SQLite file.' });
   });
 
+  it('answers the recipient\'s open question even when the client names another exchange', async () => {
+    // The dashboard sees a window over a ring buffer, so a question older than
+    // that window is invisible to it and it names the newest exchange instead.
+    // Filing that as a note would leave the agent parked forever (AGT-4030).
+    const store = await boardAt('events.json');
+    const question = await store.publish({
+      repository: '/repo', taskId: 't-park', actor: 'sable', actorName: 'Sable', actorRole: 'worker',
+      kind: 'human-question', status: 'waiting', summary: 'uv is missing — install it or run elsewhere?',
+    });
+    const laterStage = await store.publish({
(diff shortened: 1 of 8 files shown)


unohee:
## 🔍 Fresh review of d180ac9 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the server-side scoped fallback, full-board settlement logic, and both chat/orchestration composer wiring. The implementation selects the newest unsettled question for the addressed agent within repository/task scope, preserves ordinary notes when no scoped question is open, and posts answers on the blocking exchange.

PR #433, opened 2026-08-28
feat(chat): let the operator hand an agent a file, not just a path (AGT-4031)

unohee:
## Why

Handing an agent material meant placing it on the host by hand and describing the path in prose — AGT-4025 did exactly that, by shell, for one repository. This makes it an action in the chat room: drop a file, and the message the agent reads carries an absolute path that already exists.

Six cgf-portal agents parked today asking the operator for files. This is the path that answers them.

## What lands

Bytes stream straight to disk rather than through `readBody`, which caps a JSON body at 1 MiB and holds it as a string — both too small for a real file and the wrong shape for one. They land under the daemon's state directory (a mounted volume, so an upload survives a container recreate) and never inside a repository, where a build could run them or a commit could carry them.

**The store is bounded**, because it shares that volume with every worktree:

| Bound | How |
|---|---|
| 64 MiB per upload | enforced as bytes arrive; the read side is torn down on refusal, so a client that ignores the 413 is not read further |
| total ceiling (2 GiB default, env-overridable) | enforced **during** streaming against live counters, not a per-request snapshot |
| an upload that does not fit | refused with **507**, never made room for by deleting recent files |

The snapshot detail matters: staggered requests each holding their own stale total would each conclude there was room. `settledBytes` and `inFlightBytes` are both read at check time and move in one step when an upload settles.

Reclamation therefore only ever spends aged files. Uploads still streaming, and those inside a grace window, are off-limits — which is what a lowered ceiling would otherwise clear out from under a message about to be sent.

**Containment**, since agents can write to the state directory too:

- the stored name is generated here; the operator's filename is metadata only and never a path component
- the root and the task directory are each checked to be real directories, not links standing in for them — and checked again after the write, so a swap that outlasts it is caught rather than silently writing the file elsewhere and reporting success
- the file is opened `O_EXCL`, which POSIX requires to fail on a symlink
- the sweeps answer from the link rather than following it, so a directory planted under the store cannot hand the TTL sweep a repository to clear out

The sweep now also runs at startup, not only in the 2 AM job — a daemon restarted each morning before then never pruned at all.

On the client, a file that uploaded keeps its path, so retrying after a failed publish reuses it instead of uploading duplicates that count against the ceiling and make the retry likelier to fail than the attempt before it. A drop that misses the room is swallowed at the document level, where the browser would otherwise navigate away and take the composer's text and every staged file with it.

## What is deliberately not here

- **No download route.** One was written and removed: nothing fetched it. It carried a file-descriptor leak on aborted downloads, a `Content-Disposition` naming files by UUID, and a traversal surface, in exchange for a capability no consumer wanted. The agent opens the path.
- **A residual race.** Node exposes no `openat`, so a directory swapped between the check and the `open` is invisible while it happens; the post-write check catches one that outlasts the write. An actor that can win that window is already writing inside the daemon's state directory as the daemon's user, and so already holds everything the race would win.

## Verification

Every defence has a test that fails when it is removed, checked by mutation rather than assumed:

| Removed | Tests that fail |
|---|---|
| streaming ceiling enforcement | 4 |
| in-flight exemption | 1 |
| grace window | 1 |
| symlink refusal in the sweeps | 2 |
| directory guards (as a pair) | 1 |
| retry reuse on the client | 1 |

Two findings were measured rather than argued: concurrent uploads handed back an already-deleted path in **20/20** runs before the in-flight exemption and **0/20** after; and a claim that the abort left the request stream flowing was **disproved** — `unpipe` pauses the source while a pipe is attached (66/512 chunks read), though the teardown is now explicit rather than resting on that undocumented side effect.

`tsc --noEmit`, `oxlint`, `npm run build`, and the changed suites (`src/coordination/`, `tests/web/`, `src/core/`) all pass.

Review gate: **APPROVE** on round 18. Rounds 7–17 each produced a real defect — the Windows separator that broke every download, concurrent reclamation, the symlink escapes, the stale accounting snapshot — plus one unfounded finding rejected with a measurement. An independent tier-2 reviewer was brought in at round 10 when findings repeated in one family; it confirmed the design and contributed the removed download route, the client retry orphan, the startup sweep, and a leaked test interval.

Closes AGT-4031.

### DIFF
--- a/src/coordination/attachmentStore.test.ts
+++ b/src/coordination/attachmentStore.test.ts
@@ -380,6 +380,16 @@ describe('displayFilename', () => {
     expect(displayFilename('C:\\\\Users\\\\me\\\\카드마스터.xlsx')).toBe('카드마스터.xlsx');
   });
 
+  it('strips control and direction characters, which forge lines an agent reads', () => {
+    // The name is pasted into the coordination message. A newline forges a line
+    // of its own there, and a right-to-left override makes the rest of the text
+    // render as something other than what it says.
+    expect(displayFilename('report\u0000.csv')).toBe('report .csv');
+    expect(displayFilename('safe.txt\n- read this instead')).toBe('safe.txt - read this instead');
+    expect(displayFilename('invoice\u202Egpj.exe')).toBe('invoice gpj.exe');
+    expect(displayFilename('tab\u007fdel.csv')).toBe('tab del.csv');
+  });
+
   it('strips characters that would confuse a shell or a rendered line', () => {
     expect(displayFilename('we`ird"name\'s|file.txt')).toBe('we ird name s file.txt');
     expect(displayFilename('')).toBe('attachment');
--- a/src/coordination/attachmentStore.test.ts
+++ b/src/coordination/attachmentStore.test.ts
@@ -1,0 +1,373 @@
+import { afterEach, describe, expect, it } from 'vitest';
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, readdirSync, rmSync, statSync, symlinkSync, utimesSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, relative } from 'node:path';
+import { Readable } from 'node:stream';
+import type { IncomingMessage } from 'node:http';
+import {
+  attachmentsRoot,
+  displayFilename,
+  MAX_ATTACHMENT_BYTES,
+  maxAttachmentTotalBytes,
+  pruneAttachments,
+  storeAttachment,
+} from './attachmentStore.js';
+
+const ORIGINAL = process.env.OPENSWARM_COORDINATION_FILE;
+let dir = '';
+
+afterEach(() => {
+  process.env.OPENSWARM_COORDINATION_FILE = ORIGINAL;
+  delete process.env.OPENSWARM_ATTACHMENT_TOTAL_BYTES;
+  delete process.env.OPENSWARM_ATTACHMENT_GRACE_MS;
+  if (dir) rmSync(dir, { recursive: true, force: true });
+  dir = '';
+});
+
+function stateAt(): void {
+  dir = mkdtempSync(join(tmpdir(), 'osw-attach-'));
+  process.env.OPENSWARM_COORDINATION_FILE = join(dir, 'coordination.json');
+}
+
+/** A request body as a stream, which is how the route hands it over. */
+function body(content: Buffer | string): IncomingMessage {
+  return Readable.from([Buffer.from(content)]) as unknown as IncomingMessage;
+}
+
+describe('attachment storage (AGT-4031)', () => {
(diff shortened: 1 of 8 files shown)


unohee:
## 🔍 Fresh review of 46819c1 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The upload path is wired through authenticated routing and has solid non-racing symlink tests, but the sweep remains exploitable through a task-directory replacement race. Do not ship deletion/reclamation over attacker-writable directory paths until the traversal is made descriptor-anchored or otherwise race-safe.

unohee:
## 🔍 Fresh review of 1dcab6e (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The endpoint, UI wiring, authorization gate, bounded streaming, and retention scheduling are present. However, total-budget accounting is not concurrency-safe: a sweep can overwrite bytes settled by another upload after its filesystem snapshot, allowing physical storage to exceed the configured ceiling.

unohee:
## 🔍 Fresh review of e139eff (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The upload endpoint is production-wired through the authenticated API route, streams to a generated file under coordination state, enforces per-file and aggregate limits, and the UI publishes returned paths only after upload. However, the TTL branch can delete an active upload, allowing `storeAttachment` to resolve with a path already removed.

unohee:
## 🔍 Fresh review of b28a65f (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The attachment flow is wired through the authorized mutation gate, raw-stream endpoint, chat composer, retention sweep, and production static-asset build. Fix filename sanitization before merge: operator-controlled control characters remain in the stored metadata and are injected into the coordination message shown to agents.

unohee:
## 🔍 Fresh review of 011da51 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The upload route, authorization gate, client wiring, and scheduled pruning are present. However, the grace protection promised for recently completed uploads does not hold for slow streams: a completed file can be reclaimed before its message is published.

unohee:
## 🔍 Fresh review of bea1003 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Checked attachment streaming, budget enforcement, route wiring/auth gate, lifecycle pruning, and chat composer behavior. Retry caching reuses an already-uploaded attachment across a potentially different active exchange, so the task-scoped storage metadata can disagree with the task receiving the message.

unohee:
## 🔍 Fresh review of c520171 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the upload route, streamed storage/admission controls, retention and total-budget reclamation wiring, daemon sweep scheduling, and chat composer upload/publish sequencing. The implementation bounds individual and aggregate storage, avoids filename/path traversal and symlink traversal during normal operations, preserves staged files on upload/publish failure, and has production callers plus focused route/UI/store coverage.

PR #434, opened 2026-08-28
fix(daemon): let an operator's answer end the wait it was answering (AGT-4033)

unohee:
## The problem, as the operator experiences it

You answer a parked agent in the dashboard. Nothing happens for two hours.

Measured on vela today, right after answering six parked cgf-portal agents — every one delivered, every one routed onto its own question's `correlationId`:

```
[HB]   Parallel mode | slots: 2 free / 2 max | running: 0
[HB] ⏰ Skipped 39 tasks in exponential backoff period
```

Answers on the board, capacity free, nothing running. From the operator's side that is indistinguishable from being ignored — the exact failure mode AGT-4026 / AGT-4029 / AGT-4030 were meant to end.

## Why

`waiting_on_operator` parks with `setRetryTime(issueId, 4, ...)`, the last entry of `BACKOFF_MINUTES = [10, 30, 60, 120]`. That backoff is *also* the resume path — the retried run finds the answer and continues — so resume latency equals the full backoff no matter how fast the human replies.

## The fix

The heartbeat asks the board whether a parked task's questions have all been answered, and brings it forward if they have. Everything else about the backoff is untouched: a question nobody answers still waits the full two hours.

**Both halves are needed, and only one of them is the filter.** The ledger refuses to claim a `RETRY_AT` row whose time has not come (`runLedger.ts:419`), so letting the task past the heartbeat alone would have selected it and then failed to claim it, every cycle. `markReady` is what actually releases it. Production runs `automationLedgerMode: 'primary'` by default, so the durable path *is* the path — a legacy-only fix would have been dead code on the daemon it was written for.

## Four things this had to get right, each of which was wrong first

| Question | Answer | What the wrong answer did |
|---|---|---|
| Which questions count? | **All** of them, not the newest | A task runs several agents; answering one worker's question released a run parked on another's, which re-asks and parks again |
| What counts as an answer? | Only a **completed** one | A failed or expired answer event released the run into a question that is still open |
| How far back to look? | **Counted in the durable trace**, not fetched | Any window — 500 or 2,000 — lets a chatty task push its own question out of view, and the caller reads "all answered" |
| Trace unavailable? | **Say no** | The in-memory board gives exactly the false "all answered" this exists to prevent |

The last one made the reach consistent end to end: the replay in `askHuman` and the answer entrypoint now read the same durable exchange, so a question the ring has evicted can still be answered, and still returns its answer instead of being asked a second time.

## Durability

The park marker lives on the task's execution state, so it survives a restart. The answer is durable — it is on the board — and a discriminator that only existed in memory would make a restart the one thing that turns a two-minute reply back into a two-hour one.

It is retired *before* the re-admission acts on it. If that write fails the task simply keeps its backoff; re-admitting on a signal that is still set would pull a run that then fails for its own reasons forward again on every heartbeat, past the backoff that exists to stop exactly that.

## Verification

Every guard has a test that fails when it is removed, checked by mutation rather than assumed:

| Removed | Tests that fail |
|---|---|
| durable count (window instead of SQL) | 1 |
| task scoping in the answer join | 1 |
| completed-answer requirement | 1 |
| fail-closed when the trace is unavailable | 1 |
| durable question resolution | 1 |
| durable answer replay | 1 |
| park-flag discriminator | 1 |
| execution-state merge (park loses status) | 2 |

The ledger integration point is pinned directly: a `RETRY_AT` row with a future time is not claimable, and is claimable after `markReady` — which is the fact the whole fix rests on.

`tsc --noEmit`, `oxlint`, and the changed suites (`src/coordination/`, `src/automation/`, `src/taskState/`) pass — 705 tests.

Two test-isolation defects surfaced on the way and are fixed here, because reading the durable trace made them load-bearing: `humanQuestions` and `coordinationTools` shared one automation database across the whole run, so they were answering each other's questions.

Review gate: **APPROVE** on round 12. Rounds 1–11 each found something real — the legacy-only fix that production never reaches, the in-memory discriminator, the actor scoping, three separate windows, the status mismatch, the unsafe fallback, and a signal cleared too late.

Closes AGT-4033.

### DIFF
--- a/src/automation/automationDbPath.ts
+++ b/src/automation/automationDbPath.ts
@@ -22,6 +22,11 @@ let configured: string | null = null;
  *
  * Resolving both through here is what makes that hold by construction, rather
  * than by every new consumer remembering to ask.
+ *
+ * Declared once per process, at the point the process learns its configuration —
+ * a process has one automation database, the same way it has one coordination
+ * board. Callers that build a runner without going through that point (tests)
+ * redirect every store together with `OPENSWARM_AUTOMATION_DB` instead.
  */
 export function setAutomationDbPath(path: string | undefined): void {
   configured = path ? resolve(path) : null;
--- a/src/automation/automationDbPath.ts
+++ b/src/automation/automationDbPath.ts
@@ -9,9 +9,32 @@
 import { homedir } from 'node:os';
 import { resolve } from 'node:path';
 
-/** Absolute path of the automation database, honouring the test override. */
+let configured: string | null = null;
+
+/**
+ * Point every automation store at the database this deployment configured.
+ *
+ * The run ledger is handed its path explicitly, but the coordination trace is
+ * not — and since the trace stopped being a pure archive and became what says
+ * whether a parked run may be re-admitted, the two have to be the same file. A
+ * deployment that relocates its automation database must take the answers with
+ * it, or the runs move and what is known about them stays behind.
+ *
+ * Resolving both through here is what makes that hold by construction, rather
+ * than by every new consumer remembering to ask.
+ */
+export function setAutomationDbPath(path: string | undefined): void {
+  configured = path ? resolve(path) : null;
+}
+
+/**
+ * Absolute path of the automation database.
+ *
+ * The environment override wins: tests redirect every automation store at once
+ * by setting it, and must not be overruled by whatever config a runner under
+ * test was constructed with.
+ */
 export function defaultAutomationDbPath(): string {
-  return process.env.OPENSWARM_AUTOMATION_DB
-    ? resolve(process.env.OPENSWARM_AUTOMATION_DB)
-    : resolve(homedir(), '.openswarm', 'automation.db');
+  if (process.env.OPENSWARM_AUTOMATION_DB) return resolve(process.env.OPENSWARM_AUTOMATION_DB);
+  return configured ?? resolve(homedir(), '.openswarm', 'automation.db');
 }
--- a/src/automation/autonomousRunner.operatorPark.test.ts
+++ b/src/automation/autonomousRunner.operatorPark.test.ts
@@ -1,0 +1,219 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import type { TaskItem } from '../orchestration/decisionEngine.js';
+import type { DurableRunCoordinator } from './durableRunCoordinator.js';
+
+vi.mock('../core/providerOverride.js', () => ({ writeProviderOverride: vi.fn() }));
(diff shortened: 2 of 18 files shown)


unohee:
## 🔍 Fresh review of 374daba (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The durable trace/counting and durable-ledger `markReady` transition correctly address ring eviction and the ledger claim gate. However, early re-admission is only armed for `task.issueId`, while the scheduler, ledger, and coordination context consistently support the canonical fallback `task.issueId || task.id`; valid tasks without `issueId` therefore remain in `RETRY_AT` until their full backoff despite an answered question.

unohee:
## 🔍 Fresh review of a85d79c (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The durable trace and task-state plumbing are wired, and the query correctly scopes completed answers by both task and correlation ID. However, the production early-readmission transition is not exercised end-to-end, and it can clear the durable park flag before the ledger transition succeeds, permanently losing the only signal that permits early retry.

unohee:
## 🔍 Fresh review of 0960147 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The durable trace and ledger promotion paths are wired, but early re-admission uses a different task identity than the human-question producer for tasks carrying both Linear identifiers. This means answered operator questions can still wait out the full backoff in normal Linear runs.

unohee:
## 🔍 Fresh review of 0960147 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The durable trace correctly prevents board-ring eviction from hiding answers, and the RETRY_AT promotion path is fenced through markReady. However, the operator-park signal remains set after a normal backoff expiry, so a later ordinary failure can be incorrectly promoted immediately by an old answered question.

unohee:
## 🔍 Fresh review of ba84841 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The early-readmission wiring is complete for ledger/off/shadow modes and the durable promotion is correctly fenced. However, the production durable execution path does not preserve `waiting_on_operator` as a distinct retry reason, so a primary-ledger run is recorded as an ordinary failure and cannot be readmitted when its answer arrives.

unohee:
## 🔍 Fresh review of df6c426 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The early-readmission state machine and fenced ledger transition are sound, including the stale-observation race test. However, the coordination trace used to determine that an answer exists is not wired to the configured automation ledger database, so configured durable-state deployments can lose answer history across restart and fail to resume promptly.

unohee:
## 🔍 Fresh review of c516bc0 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The ledger-side fencing and backoff tests cover the intended re-admission flow. However, the new process-global database-path setting does not rebind an already-open coordination trace, so the claimed ledger/answer co-location is not guaranteed.

unohee:
## 🔍 Fresh review of f039081 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The ledger CAS/readmission logic and answered-question fail-closed behavior are correctly wired and covered. However, the process-global automation DB path makes concurrently constructed runners redirect each other’s coordination trace, violating the claimed ledger/trace co-location.

PR #437, opened 2026-08-28
fix(daemon): stop a repeatedly-unanswered ask_human from spinning (AGT-4042)

unohee:
Two things kept a parked task spinning on nothing.

## 1. Every re-dispatch re-paged the operator for the same blocker

A re-dispatched task is a fresh worker session that writes its own `ask_human` call, and it rephrases the same blocker every time. That text was hashed verbatim into the paging gate's correlation ID, so a reworded repeat minted a fresh one and defeated "page once per open question."

Live evidence today: nine Discord pages for one unanswerable ask on cgf-portal's AX-1016, thirteen tasks on the repo doing the same thing, 41 parked attempts total (`SELECT issue_id, COUNT(*) FROM automation_attempts WHERE error_code='waiting_on_operator' GROUP BY issue_id`).

Paging is now gated per `(repository, taskId)`, not per exact wording (`coordinationStore.openQuestionCount`).

## 2. Nothing stopped the automatic retry loop when the operator couldn't answer

Fixing (1) stops the spam, but the daemon still retried on a fixed ~2h clock forever, even when the missing input (e.g. a Google service-account path) simply doesn't exist in the mounted environment. Past the first unanswered ask, the run now stops entirely — `NEEDS_HUMAN` — instead of paying for another worker+reviewer cycle every backoff tick, and resumes the moment the board shows an answer, independent of Linear state.

Durable mode only: without a ledger there is nowhere to record a stop that survives a restart, so legacy mode keeps the existing backoff ladder.

## Two things this design has to hold, each pinned by a mutation-tested case

- **The operator can only ever reply to the one correlation ID they were actually paged for** — every rephrased repeat mints its own, unpaged, ID. `answerHumanQuestion` now settles every open sibling for the task when one is answered, or the open-question count this rests on never reaches zero and the park never lifts.
- **The resume check reads the run's own ledger row (`RunRecord.projectPath`) for its project path, not the in-memory path cache** — the cache is empty after a restart until something else repopulates it, and an answer that arrived while the daemon was down has to be found on the very first heartbeat back.

Both were caught by `openswarm review` on the first pass and fixed before this PR; see commit history for the isolated fixes.

Closes AGT-4042.

### DIFF
--- a/src/automation/autonomousRunner.operatorQuestionPark.test.ts
+++ b/src/automation/autonomousRunner.operatorQuestionPark.test.ts
@@ -1,0 +1,174 @@
+// ============================================
+// OpenSwarm - stop re-dispatching on a repeated, unanswered ask_human (AGT-4042)
+// ============================================
+//
+// A re-dispatch is a fresh worker session, and it is not free: burning a
+// worker+reviewer cycle every backoff tick on a question the operator has
+// already been paged for once, and has not answered, is exactly the wasted
+// spin these tests pin closed.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import type { TaskItem } from '../orchestration/decisionEngine.js';
+import type { DurableRunCoordinator } from './durableRunCoordinator.js';
+import type { PipelineResult } from '../agents/pairPipeline.js';
+import type { TaskScheduler } from '../orchestration/taskScheduler.js';
+
+vi.mock('../core/providerOverride.js', () => ({ writeProviderOverride: vi.fn() }));
+vi.mock('../agents/stageModelResolver.js', () => ({ resolveAdapterDefaultModel: vi.fn(async () => 'model') }));
+
+type InternalRunner = {
+  scheduler: TaskScheduler;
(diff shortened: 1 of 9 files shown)


unohee:
## 🔍 Fresh review of 495cbcc (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The new park is immediately undone for normal Todo tasks, and exact repeated `ask_human` calls never reach the distinct-open-question threshold. The added tests mask both paths by using Backlog for the resume case and only modeling differently worded questions.

unohee:
## 🔍 Fresh review of 0b33db7 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The repeat-attempt count is wired to the durable ledger and the NEEDS_HUMAN resume distinction is covered. However, the answer fan-out that makes rephrased asks resumable only scans the retained coordination board, so older sibling questions remain unanswered in the durable trace and permanently block the parked run.

unohee:
## 🔍 Fresh review of 3f98779 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the scheduler park/resume path, ledger streak boundary, durable trace query behavior, and human-question answer fan-out. The implementation correctly parks after two consecutive `waiting_on_operator` attempts, bounds the streak at the latest completed answer, and makes operator-question parks resume only after all durable outstanding questions are settled.

PR #439, opened 2026-08-28
fix(deps): pin adm-zip and sharp past their high-severity advisories (AGT-4047)

unohee:
## Summary
- Pin `adm-zip` (0.5.18 → 0.6.0, GHSA crafted-ZIP 4GB allocation) and `sharp` (0.34.5 → 0.35.4, inherited libvips CVEs) via `package.json` `overrides` — neither is a direct dependency; both come in transitively via `@huggingface/transformers`.
- Verified the memory embedding path end-to-end after the bump: `embedQuery` and `searchMemorySafe` both ran successfully against the real Xenova model and LanceDB table.
- `desktop/Cargo.lock`'s `glib` alert (also part of this issue) is deferred separately — see the commit message and the Linear issue comment for why.

## Test plan
- [x] `npm run build` clean
- [x] Real smoke test: `embedQuery('...')` → 768-dim vector
- [x] Real smoke test: `searchMemorySafe('...')` → `{success: true, memories: [...]}` against the live memory table
- [x] `npm audit` shows 0 open findings for `adm-zip`/`sharp` (one unrelated pre-existing `nanoid`/postcss finding split out to AGT-4049)

Fixes AGT-4047.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -468,6 +468,7 @@
       "version": "1.10.0",
       "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
       "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
+      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 2facba8 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the resolved Sharp 0.35.4 package and platform artifacts in package-lock.json against the root Node >=22 engine and node:22 Docker stages. The override is correctly reflected in the lockfile; no material compatibility or wiring defect found.

PR #441, opened 2026-08-28
fix(linear): converge a decomposition retry on its own stable sub-issue IDs (AGT-4048)

unohee:
## Summary
- `decompositionChildId(parentIssueId, index)` is deliberately content-independent — a stable hash of only `(parentIssueId, index)` — specifically so a retried decomposition converges on the sub-issues it already created instead of minting a second set.
- `createSubIssue`'s recovery path additionally required the retry's freshly re-planned title/description to match the first attempt's byte-for-byte. Planner output isn't stable across calls, so this essentially never matched — every retry burned an LLM call, hit a Linear API conflict on all 6 sub-issues, logged "existing artifact does not match the requested plan," and scheduled an identical retry. Observed live on `cgf-portal`/vela, task AX-870, 6 consecutive occurrences.
- The ID + parent match is the actual guarantee the design needs. Dropped the title/description equality from the acceptance condition; a content difference is now a diagnostic `console.warn`, not a reason to fail.

## Scope note
The issue's proposed fix also floated short-circuiting *before* the planner call when the parent already has children under these stable IDs (saving the wasted LLM call, not just the wasted Linear API calls). Left that out of this PR — it touches the decomposition/planner orchestration in a different file and is a separate, larger change; this PR is scoped to the actual root-cause bug (the recovery check itself).

## Test plan
- [x] New test: recovery succeeds when the regenerated title/description differs from the original
- [x] New test: recovery still correctly rejects when the existing artifact belongs to a *different* parent (guards against ID collision across decompositions)
- [x] `npx vitest run src/linear/linear.test.ts` — 17/17 pass
- [ ] AX-870 (or an equivalent live retry) completes decomposition instead of looping — requires this fix deployed to vela and observed against a real retry cycle; will confirm after merge + redeploy

Fixes AGT-4048.

### DIFF
--- a/src/automation/runnerExecution.coverage.test.ts
+++ b/src/automation/runnerExecution.coverage.test.ts
@@ -509,19 +509,11 @@ describe('runnerExecution.ts coverage extension', () => {
     it('falls through to direct execution when the planner says the task fits the threshold', async () => {
       plannerNeedsDecomposition.mockReturnValue(true);
       plannerRunPlanner.mockResolvedValue({
-        success: true,
-        originalIssue: 'issue-1',
-        needsDecomposition: false,
-        subTasks: [],
-        totalEstimatedMinutes: 0,
+        success: true, originalIssue: 'issue-1', needsDecomposition: false, subTasks: [], totalEstimatedMinutes: 0,
       });
       createPipelineFromConfig.mockReturnValue(makeFakePipeline(pipelineResult()));
 
-      const result = await executePipeline(
-        makeCtx({ enableDecomposition: true }),
-        task(),
-        '/repo',
-      );
+      const result = await executePipeline(makeCtx({ enableDecomposition: true }), task(), '/repo');
 
       expect(result.finalStatus).toBe('approved');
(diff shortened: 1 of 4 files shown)


unohee:
## 🔍 Fresh review of 18e1b81 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The parent-only convergence check fixes the direct Linear collision, but a re-planned retry now mixes persisted child titles with fresh dependency titles. This silently drops inter-child dependencies and can schedule blocked work before prerequisites; preserve stable slot-to-child mapping for dependency resolution or retain/reconcile the original plan.

unohee:
## 🔍 Fresh review of c69963c (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the retry recovery path in `src/linear/linear.ts:1450-1484` and dependency remapping in `src/automation/runnerExecution.ts:440-470`. Stable per-slot IDs are parent-scoped, recovered artifacts are still parent-verified, and dependencies now resolve against the current validated unique plan titles rather than stale recovered titles; focused regression tests cover both behaviors.

PR #442, opened 2026-08-28
fix(linear): converge addComment's idempotent recovery on id+issue alone (AGT-4051)

unohee:
## Summary
- Same defect shape as AGT-4048, one call deeper: `addComment`'s idempotent-collision recovery required an exact body match to converge on an already-posted comment. `buildTaskStateSyncComment` and the worker start/complete audit comments all bake a timestamp (and often a session id) into the body — that can never match a prior call's, by construction, so this failed on *every* retry that hit an already-posted stable-id comment, not just occasionally.
- Observed live: `AX-870` (the same task AGT-4048 used as its own example) hit `Conflict on insert of Comment - Entity Comment with id ... already exists.` posting its decomposition child-state-sync comment, failing the whole task.
- Fix: the stable `commentId` + owning issue is the actual identity guarantee. A body mismatch is now a diagnostic `console.warn`, not a reason to fail — mirrors AGT-4048's `createSubIssue` fix exactly.

## Test plan
- [x] New test: recovery succeeds when the existing comment's body differs (a timestamp changed)
- [x] New test: recovery still correctly rejects when the existing comment belongs to a *different* issue
- [x] Verified the fix matters: reverted the one changed condition, confirmed the new test fails exactly as predicted, restored it
- [x] `npx vitest run src/linear/linear.test.ts` — 19/19 pass

Fixes AGT-4051.

### DIFF
--- a/src/linear/linear.test.ts
+++ b/src/linear/linear.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from 'vitest';
-import { createSubIssue, drainLinearConnection, effectCommentId, fetchIssuesForStates, initLinear, parseBlockerIdentifiers } from './linear.js';
+import { addComment, createSubIssue, drainLinearConnection, effectCommentId, fetchIssuesForStates, initLinear, parseBlockerIdentifiers } from './linear.js';
 import { LinearClient } from '@linear/sdk';
 
 // createSubIssue reads the module-level client singleton (getClient()), set only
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of db69fed (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the committed diff and production call sites. `addComment` now converges only when the deterministic comment ID resolves to the target issue, while preserving the original create error for a different issue or failed reconciliation; the new tests cover both paths with divergent bodies.

PR #443, opened 2026-08-28
fix(automation): free NEEDS_RECONCILE rows a container-restart orphans forever (AGT-4052)

unohee:
## Summary
- `DurableRunCoordinator.reconcile()` only cleared a `NEEDS_RECONCILE` row's stale owner once `processIsAlive(ownerProcessId(...))` reported it dead. This container assigns the daemon the same pid every restart, so an orphaned row reads as "alive" forever — the new daemon's own pid probe hits itself.
- Same trap already fixed once in `taskState/store.ts` (`LOCK_ABANDON_MS`); this is a second, independent occurrence in `durableRunCoordinator.ts`, confirmed live on vela (3 CGF-Portal issues stuck 80+ min post-redeploy).
- Adds a time-based abandonment fallback to the `NEEDS_RECONCILE` loop only — the `CLAIMED/EXECUTING/VERIFYING/PUBLISHING` loop is untouched, since a stuck claim there is still swept by natural lease expiry.

## Test plan
- [x] New regression test proves the fix: reverted the changed line, confirmed the test failed as predicted; restored, confirmed it passes.
- [x] `npx tsc --noEmit` passes.
- [x] `npx vitest run src/automation/durableRunCoordinator.test.ts` — 31/31 passing.
- [x] Commit-gate review (tier-1, 4 rounds, each round fixed a real finding: leaseMs-derived default, explicit safety-margin justification, input validation on `reconcileAbandonMs`) — final: APPROVE.

### DIFF
--- a/src/automation/durableRunCoordinator.test.ts
+++ b/src/automation/durableRunCoordinator.test.ts
@@ -784,4 +784,63 @@ describe('DurableRunCoordinator', () => {
     deadReplacement.close();
     deadLedger.close();
   });
+
+  it('rejects a negative reconcileAbandonMs instead of silently collapsing the safety margin', () => {
+    expect(() => new DurableRunCoordinator({
+      mode: 'primary', dbPath: dbPath(), reconcileAbandonMs: -1,
+    })).toThrow(/reconcileAbandonMs/);
+  });
+
+  // AGT-4052: a container assigns the daemon the same pid every restart, so
+  // a NEEDS_RECONCILE row orphaned by a restart has processIsAlive report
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 27f3565 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the reconciliation path and production callers: `reconcile()` runs at autonomous-runner startup and each heartbeat, and age-based release remains fenced by the full ownership tuple in `RunLedger.confirmExecutorExit`. The new tests cover negative configuration, threshold boundary behavior, and the PID-reuse case where the liveness probe remains true.

PR #444, opened 2026-08-28
fix(worktree): stop a container-restart-orphaned active marker blocking recovery/prune forever (AGT-4053)

unohee:
## Summary
- `inspectWorktreeRecovery` and `pruneWorktrees` both call `processAppearsAlive(marker.ownerPid)` with no age fallback. This container assigns the daemon the same pid every restart, so a marker from a dead prior generation reads "alive" forever — third occurrence of the trap already fixed in `taskState/store.ts` and in `DurableRunCoordinator.reconcile()` (AGT-4052, #443).
- Confirmed live on vela: AX-1018/AX-885/AX-1044's active-worktree markers all carry `ownerPid: 7` (the daemon's fixed container pid) — without this fix, these 3 issues would stay blocked even after AGT-4052 ships.
- Adds a time-based abandonment fallback (default 24h, overridable via `OPENSWARM_ACTIVE_MARKER_ABANDON_MS`) to both call sites.

## Test plan
- [x] New regression tests prove the fix: reverted, confirmed failures as predicted; restored, confirmed passing.
- [x] `npx tsc --noEmit` passes.
- [x] `npx vitest run src/support/worktreeManager.coverage.test.ts src/support/worktreeManager.test.ts src/support/worktreeManager.auditPR.test.ts` — 78/78 passing.
- [x] Commit-gate review (tier-1, 3 rounds — each fixed a real finding: initially scoped `pruneWorktrees` too, verified via `stageTimeoutMs`/`maxIterations` math it's actually safe there too; then caught my comment overclaiming a "hard ceiling" that configured stage timeouts aren't actually capped by, fixed by making the threshold overridable via env var) — final: APPROVE.

### DIFF
--- a/src/support/worktreeManager.coverage.test.ts
+++ b/src/support/worktreeManager.coverage.test.ts
@@ -244,6 +244,48 @@ describe('createWorktree retry/resume error paths', () => {
     expect(readFileSync(join(worktreePath, 'stray.txt'), 'utf8')).toBe('leftover');
   });
 
+  // AGT-4053: this container assigns the daemon the same pid every restart,
+  // so a marker from a dead prior generation makes processAppearsAlive hit
+  // the *current* process's own pid and read "alive" forever. Age must be
+  // able to reclassify it as orphaned even while the pid probe insists the
+  // marker's owner is alive.
+  it('reclassifies a very old active marker as orphaned even when its pid still appears alive (AGT-4053)', async () => {
+    const info = await createWorktree(repo, 'INT-1', 'swarm/INT-1-pid-reuse');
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of f28df27 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Reviewed the recovery and prune call paths plus their production reconciler wiring. The new age gate is applied consistently before treating a PID-marked owner as live, malformed timestamps remain fail-closed, and the integration tests exercise real worktrees with both the default and environment override thresholds.

PR #445, opened 2026-08-29
feat(coordination): nudge worker/reviewer agents to check their inbox mid-task (AGT-4054)

unohee:
## Summary

- An operator's unsolicited message to a running worker (AX-858, CGF-Portal) landed correctly on the coordination board, but the worker never called `coordination_read` to see it — nothing actively prompts a running agent to check its pull-based inbox.
- Add a periodic in-loop nudge in `agenticLoop.ts`: every 6 turns without a real inbox check, inject a `coordination_read` reminder. Only a successful `coordination_read` resets the clock — `coordination_history` searches the permanent trace and consumes nothing live, so it must not count (caught by round-1 `openswarm review`, fixed + regression-tested).
- Pair the nudge with short static system-prompt guidance on worker/reviewer (`COORDINATION_GUIDANCE_PROMPT`) explaining what to do once a message is found: acknowledge via `coordination_publish` before finishing the stage, not silently fold it into the next edit.
- Fix `coordination_history`'s dispatcher, which fell through to `"Unknown tool"` for every call — independently discovered while researching this feature, worth shipping in the same change since a robust "catch up on messages" flow wants this tool working.

**Scope note (adapter type)**: this only benefits the 6 adapters that run through OpenSwarm's own turn loop (`codex-responses` — this deployment's default — `cc-router`, `gpt`, `openrouter`, `atlascloud`, `local`). The other 4 (`claude`, `codex`, `cursor`, +1) shell out to an external CLI's own opaque tool loop with zero turn-level visibility, and already drop coordination tools entirely (`base.ts:88-101`) — unrelated to and not fixed by this change.

**Scope note (execution path)**: this benefits any run that already has a `coordinationContext` wired in — confirmed for the autonomous daemon pipeline (`pairPipeline.ts`'s `coordinationContextFor`, called for both worker and reviewer). The interactive Discord `/pair` flow (`discordPair.ts`) calls `worker.runWorker`/`reviewer.runReviewer` directly and never constructs a `coordinationContext` at all, so this PR's nudge/guidance/tools are inert there today — verified via the fresh PR review, confirmed by reading `discordPair.ts` directly. Wiring that up isn't a small addition: `coordinationContextFor` needs a `TaskItem`-shaped `{id, issueId}` (`taskEventKey`), while a `PairSession` only carries a flat `taskId` string, and a Discord-only pair session doesn't always originate from a Linear issue — a real design question, not touched here. Tracked as a follow-up.

## Test plan

- [x] `npx vitest run src/adapters/agenticLoop.test.ts src/adapters/tools.test.ts src/agents/worker.test.ts src/agents/reviewer.test.ts` — 167/167 passed
- [x] Each new test individually verified to fail when its corresponding fix is reverted (nudge injection, reset detection, `coordination_history` exclusion, dispatcher fix)
- [x] `openswarm review` round 1: REVISE (`coordination_history` wrongly reset the check clock) → fixed + regression test added
- [x] `openswarm review` round 2: **APPROVE**
- [x] `openswarm pr review --fresh`: REVISE (Discord pair path has no coordination context) → verified true, out of this PR's scope, documented above, follow-up filed

### DIFF
--- a/src/adapters/agenticLoop.test.ts
+++ b/src/adapters/agenticLoop.test.ts
@@ -10,7 +10,7 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { promises as fs } from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
-import { compactPriorTurns, toolCallKey, allToolCallsSeen, shouldNudgeReadLoop, READ_LOOP_NUDGE_AT, runAgenticLoop, loopResultToCliResult, type ChatMessage, type AgenticLoopResult } from './agenticLoop.js';
+import { compactPriorTurns, toolCallKey, allToolCallsSeen, shouldNudgeReadLoop, READ_LOOP_NUDGE_AT, shouldNudgeCoordinationCheck, COORDINATION_CHECK_NUDGE_EVERY, runAgenticLoop, loopResultToCliResult, type ChatMessage, type AgenticLoopResult } from './agenticLoop.js';
 import type { ToolCall } from './tools.js';
 
 /** Scripted API response carrying a single tool call. */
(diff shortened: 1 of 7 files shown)


unohee:
## 🔍 Fresh review of 24c0beb (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The loop, dispatcher, guidance prompt, and unit-level nudge behavior are internally consistent. However, the primary Discord pair execution path never supplies a coordination context to either worker or reviewer, so the new tools, guidance, and inbox nudges are disabled in that production workflow.

unohee:
Filed AGT-4055 to track the Discord /pair coordinationContext gap the fresh review surfaced — deferred out of this PR's scope, see PR body for why.

PR #446, opened 2026-08-29
fix(automation): reopen branchless orphaned NEEDS_RECONCILE rows instead of leaving them stuck forever (AGT-4056)

unohee:
## Summary

- AGT-4052/4053 cleared a NEEDS_RECONCILE row's stale owner/lease once a container-restart pid-reuse trap could no longer disprove it by age, but nothing ever advanced the row's **state** afterward. `NEEDS_RECONCILE` is not in `CLAIMABLE_STATES` (can never be reclaimed) yet `claimRun()`'s admission check counts it toward the per-project active-run cap — so a row that reaches this state permanently squats one admission slot, forever.
- Live evidence, found during 24h autonomous-loop monitoring: 3 CGF-Portal rows (AX-1018, AX-885, AX-1044) sat 2+ hours with owner/lease already null and `pr_url` null; cgf-portal's `maxConcurrentTasks=2` meant these 3 zombie rows alone permanently blocked every other claim in the project, catching 5 issues (AX-863, AX-872, AX-1027, AX-1028, AX-1030) in an infinite `claim_deferred` retry loop, `attempt_no` climbing into double digits.
- `reconcile()`'s NEEDS_RECONCILE loop now reopens an orphaned row straight to READY when it has **no `branchName` at all** (nothing was ever pushed, so there's nothing to look up).
- A row **with** a `branchName` is left alone: `prUrl == null` only proves the ledger never recorded a PR, not that one was never published — `reconcileDurableArtifacts()` already owns that case via a GitHub-by-branch lookup before falling back to worktree-evidence inspection, and reopening here would race ahead of it and risk duplicating published work. **This scope narrowing was caught by round-1 `openswarm review`, not self-caught** — my first draft used `prUrl == null` alone.

## Important: does not resolve AX-1018/885/1044 directly

All 3 production rows carry a `branchName`, so this fix's condition does not apply to them. Investigating further: they're each on `worktreeManager.ts`'s own, already-shipped (AGT-4053) `activeMarkerAbandonMs` 24h timer — a deliberate safety margin (the worktree marker has no per-heartbeat renewal to lean on, unlike a ledger lease), not a bug. Their markers were 4h47m-21h50m old at check time; they'll self-heal on that timeline. See AGT-4056 for the full correction and the related throughput-vs-safety tradeoff this surfaced (not changed here — flagged as an open question).

## Test plan

- [x] `npx vitest run src/automation/durableRunCoordinator.test.ts src/automation/runLedger.test.ts src/automation/autonomousRunner.durable.test.ts` — 98/98 passed
- [x] Each new/modified test individually verified to fail when its corresponding fix piece is reverted
- [x] `openswarm review` round 1: REVISE (`prUrl == null` alone isn't proof nothing was published for a branch-bearing row) → fixed + regression test added
- [x] `openswarm review` round 2: **APPROVE**

### DIFF
--- a/src/automation/durableRunCoordinator.test.ts
+++ b/src/automation/durableRunCoordinator.test.ts
@@ -200,9 +200,14 @@ describe('DurableRunCoordinator', () => {
     ownershipProbe.mockRestore();
     exitAck.mockRestore();
     warning.mockRestore();
-    expect(coordinator.reconcile(Date.now() + 4_000)).toHaveLength(1);
+    // 2, not 1: the expired-lease reconciliation itself, plus the same call's
+    // NEEDS_RECONCILE pass auto-reopening the row straight to READY once its
+    // owner/lease just cleared and nothing was ever published (AGT-4054
+    // follow-up — previously this sat in NEEDS_RECONCILE forever, unclaimable
+    // yet still counted against the project's admission cap).
+    expect(coordinator.reconcile(Date.now() + 4_000)).toHaveLength(2);
     expect(ledger.getRun('LEDGER-READ-FAIL')).toMatchObject({
-      state: 'NEEDS_RECONCILE',
+      state: 'READY',
       ownerInstanceId: undefined,
       leaseToken: undefined,
     });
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 93cadea (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the coordinator’s expired-lease flow, `RunLedger.markReady()` fencing, and `AutonomousRunner.reconcileDurableArtifacts()` ordering. The new path only reopens ownerless rows with neither durable branch nor PR evidence; branch-bearing and published rows remain parked for artifact reconciliation, and the tests cover each distinction.

PR #447, opened 2026-08-29
feat(cli): add `openswarm attach` to upload files to a task's coordination inbox (AGT-4058)

unohee:
## Summary

Neither `POST /api/coordination/attachment` (AGT-4031) nor `POST /api/coordination/message` had a CLI consumer — only the dashboard's `/chat` page could hand a running agent a file. This adds `openswarm attach <issueId> <files...> [-m/--message]`, following the dashboard's own flow: upload the file(s), fold their returned paths into a coordination message, address it to whichever agent last spoke on the task (preferring an open blocking question over a plain note, matching AGT-4030's precedence).

- `taskId` on a coordination event is the Linear issue **UUID**, not the `AGT-123`-style identifier, so the command resolves that the same way `openswarm work` does (`ensureTaskSource` + `getIssue`).
- `latestAddressable`/`openQuestionFor` are ported natively from `web/static/js/conversationModel.mjs` (small, pure functions) rather than imported across the `src/`→`dist/` / `web/static/`→`dist/web-static/` build boundary, which resolves differently in dev vs. prod.

## Two defects caught by review, not self-caught

1. **`latestAddressable` could address a daemon.** It had to additionally exclude `actorRole: 'daemon'` beyond what the dashboard's own version excludes (just `'human'`) — `adapter-route` (`worker.ts`'s `recordRoute`, actor `adapter-router`) and `mcp-audit` (`daemonActor` in `runCoordination.ts`) are both daemon-authored bookkeeping; neither identity ever runs an agentic loop or calls `coordination_read`, so addressing one would silently strand the message forever. The CLI has no human in the loop to notice a bad recipient pick the way a dashboard operator would. Filed **AGT-4059** to bring the dashboard's own copy in line (lower priority there, since a human is watching).
2. **Every daemon request after the health probe was unbounded** and could hang the CLI indefinitely against a responsive-but-stalled daemon. Added `AbortSignal.timeout` to all of them (10s for JSON calls, 60s for uploads — a file can be up to `MAX_ATTACHMENT_BYTES`, 64 MiB).

Both fixed with regression tests.

## Test plan

- [x] `npx vitest run src/cli/attachHandler.test.ts src/cli/providerCommand.test.ts src/cli/workCommand.test.ts` — 82/82 passed
- [x] Each new test individually verified to fail when its corresponding fix is reverted (question-preference, no-addressable-agent guard, daemon-unreachable guard, daemon-actor exclusion, request timeouts)
- [x] Ported `latestAddressable`/`openQuestionFor` covered by the same scenarios as `tests/web/conversationModel.test.ts`, proving the TS port matches the dashboard original
- [x] `openswarm review` round 1: REVISE (daemon-actor recipient) → fixed
- [x] `openswarm review` round 2: REVISE (unbounded requests) → fixed
- [x] `openswarm review` round 3: **APPROVE**

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -722,6 +722,19 @@ program
     process.exitCode = await runProviderCommand(name);
   });
 
+// openswarm attach
+
+program
+  .command('attach')
+  .description("Upload file(s) to a task's coordination inbox and notify its agent")
+  .argument('<issueId>', 'Linear issue id or identifier (e.g. AGT-123)')
+  .argument('<files...>', 'Local file path(s) to upload')
+  .option('-m, --message <text>', 'Message to send with the file(s)')
+  .action(async (issueId: string, files: string[], opts: { message?: string }) => {
+    const { runAttachCommand } = await import('./cli/attachHandler.js');
+    process.exitCode = await runAttachCommand(issueId, files, opts);
+  });
+
 // openswarm dash
 
 program
--- a/src/cli/attachHandler.test.ts
+++ b/src/cli/attachHandler.test.ts
@@ -1,0 +1,305 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { CoordinationEvent } from '../coordination/coordinationStore.js';
+import {
+  latestAddressable,
+  openQuestionFor,
+  resolveIssue,
+  runAttachCommand,
+} from './attachHandler.js';
+
+const fetchMock = vi.fn();
+vi.stubGlobal('fetch', fetchMock);
+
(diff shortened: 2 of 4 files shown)


unohee:
## 🔍 Fresh review of cd47d00 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked CLI registration, issue resolution, daemon authentication/wiring, coordination route contracts, trace ordering, and attachment persistence. The command resolves the Linear UUID before querying/uploading, bounds every daemon request, refuses uploads without an addressable agent, and sends the notification using a pending human-question correlation when applicable.

PR #448, opened 2026-08-29
fix(coordination): stop publishing "(no summary)" and bare codenames as an agent's report (AGT-4060)

unohee:
## Summary

The user reported the coordination board showing lines like `Nimbus (worker · AX-859) → reviewer-eb2a: (no summary)`. Confirmed live on vela's `coordination_trace`: every degraded line is a `delegation-result` whose summary is either the literal `(no summary)` or a bare `Codename: Atlas` / `Codename: 린터`. Two independent causes, both defeating the same existing fallback.

`publishStageOutcomeToBoard` already falls back to a timing line when an agent said nothing (`said || (success ? 'Finished in Ns' : 'Did not pass in Ns')`) — but two non-reports reach it as **truthy strings**, so `||` takes them:

1. **`(no summary)`** — adapters substitute this literal for a missing summary (`claude.ts:213`, and the same pattern in `documenter.ts`, `auditor.ts`, `skillDocumenter.ts`). The board showed a placeholder where a duration would at least have been true.
2. **`Codename: X`** — `worker.ts:481` strips the agent's self-introduction (AGT-4019) but with a `|| parsedResult.summary` restore, so a summary that was *only* that line comes back intact and reads as the agent reporting its own name as its work.

Fixed with a `boardWords()` helper that normalizes both to `undefined` at the board boundary, so the existing fallback fires. Fixing it here rather than in the adapters leaves other consumers of `(no summary)` untouched; `src/linear/format.ts:95-101` already strips the codename line without restoring for the Linear-comment path, so this applies the same rule to the board.

## Live evidence (vela, before the fix)

| kind | from → to | summary |
|---|---|---|
| delegation-result | Delta → reviewer-a4f2 | `(no summary)` |
| delegation-result | Atlas → reviewer-8eb5 | `Codename: Atlas` |
| delegation-result | 린터 → reviewer-167f | `Codename: 린터` |
| advice-response | worker-6c1f → reviewer-c1ef | (healthy — real sentence) |
| delegation-request | worker-6c1f → reviewer-c1ef | (healthy — `Taking on: ...`) |

Only `delegation-result` degraded; the other kinds were unaffected.

## Test plan

- [x] `npx vitest run src/agents/pairPipeline.board.test.ts src/agents/worker.test.ts` — 52/52 passed
- [x] 5 new cases: `(no summary)` → timing fallback; codename-only → timing fallback; failure wording on a failed reportless stage; a **real summary passes through untouched**; a codename line *preceding* a real report keeps the report
- [x] Each new test verified to fail when `boardWords` is reverted (4 of 5 fail; the real-summary case correctly passes either way)
- [x] `openswarm review`: **APPROVE**

Not yet done (post-merge): confirm on vela after redeploy that new `delegation-result` events carry real text.

### DIFF
--- a/src/agents/pairPipeline.board.test.ts
+++ b/src/agents/pairPipeline.board.test.ts
@@ -86,6 +86,46 @@ describe('stage lifecycle on the coordination board', () => {
     expect(event).toMatchObject({ actorName: 'Sable', recipientRole: 'worker', status: 'failed', correlationId: exchangeId });
   });
 
+  // AGT-4060: the board showed `(no summary)` and bare `Codename: X` lines
+  // where the agent's report should be. Both reached it as truthy strings, so
+  // the `said || timing-fallback` never fired. Reported by the user watching
+  // the live board, not caught by a test.
+  describe('a summary that carries no report falls through to the timing fallback (AGT-4060)', () => {
+    async function outcomeSummary(summary: string | undefined, success = true): Promise<unknown> {
+      const ctx = context({ task: { id: `t-${summary ?? 'none'}`, issueId: `i-${summary ?? 'none'}`, issueIdentifier: 'AGT-60', title: 'T' } });
+      publishStageOutcomeToBoard(ctx as never, 'worker', {
+        success, durationMs: 3_300, result: { summary },
+      }, stageCorrelationId(ctx as never, 'worker'));
+      await new Promise((resolve) => setTimeout(resolve, 0));
+      await new Promise((resolve) => setTimeout(resolve, 0));
+      return (published.events.at(-1) as Record<string, unknown>).summary;
+    }
+
+    it("does not publish the adapter's literal `(no summary)` placeholder", async () => {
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 7043345 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The English placeholder and codename-only cases are correctly routed to the duration fallback, and the board call path is wired through `pairPipeline.ts`. However, the placeholder check is locale-specific while adapters use the active locale's `common.fallback.noSummary`; Korean deployments still publish `(요약 없음)` as the report.

unohee:
## 🔍 Fresh review of cf53a02 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the committed diff, production call site, locale state, and existing worker handling. `boardWords()` correctly converts both active-locale placeholders and codename-only summaries to the existing duration fallback while preserving substantive text; the added tests cover success, failure, multiline, and Korean-locale behavior.

PR #449, opened 2026-08-29
fix(tools): let agents read local material linked in from the main checkout (AGT-4061)

unohee:
Closes AGT-4061.

## The bug

A repo can hand its worktrees material git cannot carry — client data, `.env` files — by symlinking it in from the main checkout on `post-checkout`. cgf-portal does exactly that for `docs/CGF_data` and the `.env` family.

`validatePath` canonicalizes through the symlink (`realpathSync`) and then requires the **target** to be inside the worktree. The target is in the main checkout, so every read was refused.

Live impact: `worker-86be` on AX-1039 asked the operator twice (06:40 and 11:36 KST) for files that were linked into its worktree and readable the whole time, then parked as `NEEDS_HUMAN`. **The agent's report was accurate, not a hallucination.** It went unnoticed because the earlier verification used a *shell* — `bash` is not path-validated, `read_file` is.

## The fix

Accept a canonical path landing inside the main checkout this worktree belongs to — for `read_file` and `search_files` only.

| | before | after |
|---|---|---|
| `read_file` / `search_files`, normal run | refused | **allowed** |
| `write_file` / `edit_file` / `apply_patch` | refused | refused |
| any tool, `readOnly` run | refused | refused |
| unrelated path outside both | refused | refused |
| `/tmp` | allowed | allowed |

**Writes keep the old containment.** Writing into the main checkout breaks worktree isolation — the exact failure `link-local-assets.sh` documents for `.venv`, where a guard test kept passing because the import resolved through the main tree.

**Read-only runs get nothing.** An ordinary worker can already reach the main checkout through the unvalidated `bash` tool, so there this is a usability fix, not a widening. A read-only reviewer has `bash` denied (`READ_ONLY_DENIED_TOOLS`), which makes this sandbox its real outbound boundary — INT-3189 — and it stays untouched.

## Why the link is checked in both directions

The main checkout is read from git's metadata, not guessed: a linked worktree's `.git` is a file holding `gitdir: <main>/.git/worktrees/<name>`.

But **that file is inside the sandbox** — `write_file` can reach it. An agent could rewrite it to `gitdir: /etc/.git/worktrees/x`, or point it at another repository's real worktree metadata on the same host, and open that checkout to reads. git also writes a back-link at `<gitDir>/gitdir` pointing back at this worktree, and *that* file lives in the main checkout where no write path reaches. Requiring both directions turns a forgeable one-way pointer into a link only git could have created.

Both links resolve against their own directory, so `worktree.useRelativePaths=true` repos (git ≥ 2.48) work — the back-link there is `../../../../wt/.git`, and resolving it against the process cwd yielded a garbage path that silently rejected every such worktree. Caught by the commit-gate review, reproduced, fixed, and covered.

Uncached deliberately: measured ~36 µs per resolve (noise beside the file read it guards), and a path-keyed cache would go stale in a daemon that outlives many worktrees.

## Verification

10 new tests in `tools.test.ts`, **each verified to fail when the behaviour it covers is reverted** (8 mutations):

| mutation | test(s) killed |
|---|---|
| exception removed entirely | symlink read; genuine back-link accepted |
| exception leaked into `write_file` | write refused |
| `readOnly` gate dropped | read-only read refused |
| widened to the main checkout's *parent* | unrelated outside path refused |
| plain checkout falls back to an ancestor | plain checkout gets nothing |
| back-link requirement removed | no-back-link refused; wrong-target refused |
| back-link existence checked but not direction | wrong-target refused |
| back-link resolved against process cwd | relative-path worktree read |

Fixtures live under `/var/tmp`, **not** `/tmp` — `validatePath` allows `/tmp` outright, so a fixture there would pass with the fix reverted.

`npm run typecheck` clean, `oxlint` clean, 185 tests green across all 9 test files that import `tools.js`.

Commit gate: `openswarm review` ×3 — APPROVE → REVISE (the relative-path back-link, above) → APPROVE.

## Still to do after merge

Verify on vela that an agent's **file tool** can read under `docs/CGF_data`. A shell check is explicitly not sufficient evidence here — that is the exact trap that hid this bug.

### DIFF
--- a/src/adapters/tools.test.ts
+++ b/src/adapters/tools.test.ts
@@ -16,6 +16,16 @@ vi.mock('../memory/repoKnowledge.js', () => ({
       : 'A non-empty query is required.',
 }));
 
+// `worktree.useRelativePaths` landed in git 2.48; older git always writes the
+// worktree back-link as an absolute path, so the relative-mode test has nothing
+// to discriminate and is skipped rather than passing vacuously.
+let hasRelativeWorktrees = false;
+try {
+  const raw = execFileSync('git', ['--version'], { stdio: 'pipe' }).toString();
+  const [, major, minor] = /(\d+)\.(\d+)/.exec(raw) ?? [];
+  hasRelativeWorktrees = Number(major) > 2 || (Number(major) === 2 && Number(minor) >= 48);
+} catch { /* git not installed */ }
+
 // Check if rg binary is available (not just a shell function wrapper)
 let hasRg = false;
 try {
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 6f35b35 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the production wiring for read_file and search_files, read-only exclusion, canonical containment, bidirectional worktree metadata verification, and the absolute/relative-link and forged-link tests. The implementation confines the exception to authenticated linked-worktree main checkouts and read-capable tools; no material defect found.

PR #450, opened 2026-08-29
fix(security-audit): bound concurrent CodeQL runs so audits stop starving the daemon (AGT-4062)

unohee:
Closes AGT-4062.

## The problem

The security-audit gate runs CodeQL **per task** — once for a baseline, once after the work — with nothing capping how many run at a time. Its cost therefore scaled with `maxConcurrentTasks` (6 on the deployed daemon) while the container's budget stayed fixed at 4 CPUs / 16 GiB.

Measured on the live daemon, 2026-08-29:

`/api/health` builds a **pure in-memory payload** (`healthEndpoint.ts` — zero I/O, version read once at module load), so its latency reads event-loop delay directly:

```
23.59  32.84  0.06  0.00  32.51  0.15  33.72  0.08  40.43  0.03  0.20  34.90  0.06  0.13   (s)
n=14   p50=0.20   over 10s: 6/14   over 25s: 5/14   max=40.43
```

Bimodal — sub-second or tens of seconds, nothing between. The signature of a thread that simply is not being scheduled.

| cgroup v2 | value |
|---|---|
| `cpu.max` | `400000 100000` → 4.0 CPUs |
| observed | **399.90%** — pinned at the cap |
| `memory.max` | 16 GiB |
| **`memory.peak`** | **13.76 GiB (86%)** |

RSS inside the container: CodeQL java **3.32 GB** + **3.21 GB**, daemon `node` **3.41 GB (1 thread)**.

Two consequences, one visible and one sharp:

- The healthcheck (`curl /api/health`, 10s timeout) hit `FailingStreak: 5` and the container read `Up 2 hours (unhealthy)` with `RestartCount: 0`. It was working fine. That is worse than noise — it stops `unhealthy` from meaning *broken*, which is the one thing a healthcheck is for.
- At 86% of the memory limit with only two audits overlapping, one more plausibly crosses 16 GiB and the kernel OOM-kills the container, **orphaning every in-flight task at once**.

## The fix

A process-wide gate held across the snapshot copy and every CodeQL invocation, plus an explicit `--ram` target per run.

At a cap of 1: ~3.3 GB CodeQL + 3.4 GB daemon ≈ 7 GB of 16, and 2 of 4 CPUs (`--threads=2`), leaving room for the daemon's loop and the agent CLIs.

Deliberately **not** lowering `maxConcurrentTasks`: that cuts all task parallelism to contain one expensive stage. Bounding the stage keeps the rest.

`--ram=<MB>` verified against the **deployed CodeQL 2.26.4** on both subcommands (`-M, --ram=<MB>`). Its own help calls analyze's version best-effort — *"may be broken by file-backed…"* — so the gate, not the flag, is what actually bounds usage. The wording in the config and type comments says so rather than claiming a ceiling.

## Why the cap is not configurable per audit

It was, briefly. Three review rounds each found a **different** defect in that shape:

| round | finding |
|---|---|
| 2 | a later caller silently raises the cap out from under a running audit |
| 3 | a cap lowered mid-flight can never take hold — each release refills the seat it vacated |
| 4 | a raise is *discarded* rather than deferred, so under sustained load the configured value never applies |

All three are one mistake: **a process-wide resource bound is not a per-call argument.** So the field is gone rather than patched. The cap is read once at module load from `OPENSWARM_CODEQL_MAX_CONCURRENT` (default 1, max 8) — the idiom this repo already uses for its other 30+ process-wide knobs — and is immutable thereafter. A missing or unparseable value falls back to the safe default rather than failing the daemon's start.

Removing mutability also removed the cap re-check in `release()`, which became provably unreachable once the limit was fixed; leaving it would have been dead code.

`maxRamMb` stays in config: it is a genuine per-run argument with no shared state.

## Cleanup cannot strand a slot

`release()` is nested in its own `finally`. `rm` runs with `force: true`, which swallows ENOENT but **not** EACCES/EPERM/EBUSY. Before the gate, a failed cleanup only leaked a temp directory; with a cap of 1 it would strand the slot and deadlock every later audit in the process. Found by the commit-gate review.

## Verification

33 tests in `securityAudit.test.ts`, **each verified to fail when the behaviour it covers is reverted**:

| mutation | test(s) killed |
|---|---|
| gate removed entirely | serialized audits; per-process bound |
| LIFO instead of FIFO | arrival order |
| env parser accepts out-of-range | 5 parser cases |
| `--ram` dropped | per-invocation RAM argument |
| cleanup throw skips release | slot returned on cleanup failure |
| `reset()` leaves the slot held | reset frees the slot |
| release underflows past zero | active count floor |

Test isolation: the gate is module state shared by every case in the file, so `reset()` runs in `beforeEach`. Without it a single leaked hold made every later case wait out its full timeout — an injected fault took **90s** and produced 5 cascading failures; with isolation and explicit 3s timeouts it fails in **6.4s** at the actual fault.

One earlier round of my own tests did **not** discriminate and was rewritten: a mocked SARIF missing `version: '2.1.0'` made audits end in `failed`, which several assertions never looked at, so they passed regardless.

`npm run typecheck` clean, `oxlint` clean, **170 tests green** across all 5 files touching `securityAudit`. Commit gate: `openswarm review` ×5 — REVISE ×3 (each a new, correct finding), then APPROVE.

## Deliberately out of scope

The healthcheck timeout. Raising it before knowing the post-fix latency distribution would be guessing at the number. It is a compose-level override, so re-tuning after measurement costs no image rebuild. Tracked in the issue's DoD.

## After merge

Re-measure on the daemon and report `/api/health` p90 and `memory.peak` as a before/after against the numbers above.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -155,6 +155,14 @@ autonomous:
   securityAudit:
     enabled: true
     maxThreads: 2
+    # Memory target passed to each CodeQL run, in MB. Best-effort by CodeQL's
+    # own account — a large database can still exceed it.
+    maxRamMb: 4096
+    # How many CodeQL runs may execute at once is a PROCESS-wide bound, not a
+    # per-audit setting, so it lives in the environment rather than here:
+    # OPENSWARM_CODEQL_MAX_CONCURRENT (default 1, max 8). Raise it only on a
+    # host with cores to spare — the audit runs twice per task, and letting it
+    # scale with maxConcurrentTasks is what starved the daemon (AGT-4062).
 
   # Pipeline guards
   guards:
--- a/src/automation/prProcessor.test.ts
+++ b/src/automation/prProcessor.test.ts
@@ -192,7 +192,7 @@ describe('PRProcessor.fixOne (INT-3282)', () => {
     await newProcessor().fixOne(pr, '/tmp/proj');
 
     const args = createPipelineFromConfigImpl.mock.calls.at(-1);
-    expect(args?.[9]).toEqual({ enabled: true, maxThreads: 2 });
+    expect(args?.[9]).toEqual({ enabled: true, maxThreads: 2, maxRamMb: 4096 });
   });
 
   it('succeeds when there are no conflicts, the pipeline succeeds, and CI passes', async () => {
--- a/src/cli/prCommand.test.ts
+++ b/src/cli/prCommand.test.ts
@@ -203,7 +203,7 @@ describe('loadRolesBestEffort / buildProcessorConfig (INT-3282)', () => {
   });
 
   it('defaults PR remediation to the enabled CodeQL policy', () => {
-    expect(loadSecurityAuditBestEffort()).toEqual({ enabled: true, maxThreads: 2 });
+    expect(loadSecurityAuditBestEffort()).toEqual({ enabled: true, maxThreads: 2, maxRamMb: 4096 });
   });
 
   it('enables the conflict resolver by default', () => {
--- a/src/core/config.test.ts
+++ b/src/core/config.test.ts
@@ -151,6 +151,7 @@ agents:
       expect(config.autonomous?.securityAudit).toEqual({
         enabled: true,
         maxThreads: 2,
+        maxRamMb: 4096,
       });
     });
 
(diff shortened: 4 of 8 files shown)


unohee:
## 🔍 Fresh review of ee5dbec (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the process-wide gate lifecycle, CodeQL `--ram` wiring for both database create/analyze, schema defaults and production propagation through autonomous and PR pipelines. The slot is released even when snapshot cleanup fails; configuration validation supplies `maxRamMb` while the runtime fallback preserves compatibility with existing callers.

PR #451, opened 2026-08-29
feat(health): report the daemon's heap so a stall can be measured, not inferred (AGT-4063)

unohee:
Part of AGT-4063 (the reporting half; the decision it enables comes after measurement).

## Why

Diagnosing the 23–40 s event-loop stalls in AGT-4062, the daemon showed:

```
VmRSS  3.55 GB      VmHWM 3.71 GB       # plateaued, not still climbing
heap_size_limit     4144 MB             # Node's default; NODE_OPTIONS empty
```

with `/api/health` latency **bimodal** — sub-second or tens of seconds, nothing between. That is the shape of stop-the-world mark-compact pauses on a heap near its ceiling, and it is a *different* cause from the CPU contention AGT-4062 addresses.

It could not be confirmed, because nothing reported the heap:

```
grep -rn "memoryUsage\|heapUsed\|getHeapStatistics" src --include='*.ts' | grep -v test
→ (no matches)
```

RSS is not heap usage. Reading `used_heap_size` through the inspector would mean signalling a production daemon — not worth it for a diagnostic.

## What

`heap_used_mb`, `heap_limit_mb`, `rss_mb` on the existing payload.

Both sources are in-process reads of counters V8 already maintains — no syscall, no I/O. That matters here specifically: the endpoint's diagnostic value is that **its latency reads event-loop delay rather than work it does itself**, and a syscall would silently destroy that.

The exhaustive `toEqual` on the vega BackendHealth contract stays exhaustive, with the three fields added, so a future addition has to be acknowledged rather than slipping into the shape that shell parses.

## Verification

Both mutations kill the right tests:

| mutation | tests killed |
|---|---|
| memory fields dropped from the payload | contract shape; MB reporting; live-read fallback |
| bytes reported instead of MB | same three |

`npm run typecheck` clean, `oxlint` clean, 10/10 in `healthEndpoint.test.ts`. Commit gate: `openswarm review` — APPROVE, first round.

**One test I wrote was removed rather than kept**: an assertion that the payload does no file I/O, via a spy on `fs.readFileSync`. It cannot fail — the module captures `readFileSync` in a local binding at load time, so a namespace spy never intercepts it, and `VERSION` is read once at load anyway. A test that cannot fail is worse than no test; the property is documented in the code comment instead.

## Sequencing

This is observability only. It ships alongside the AGT-4062 gate deliberately: reporting a number does not change behaviour, so it cannot confound the measurement of whether that gate fixes the stalls.

### DIFF
--- a/src/support/healthEndpoint.test.ts
+++ b/src/support/healthEndpoint.test.ts
@@ -37,7 +37,10 @@ describe('buildHealthPayload', () => {
       uptimeS: 12.9,
       version: '1.2.3',
       instanceId: 'fixed-id',
+      memory: { heapUsedBytes: 1024 * 1024, heapLimitBytes: 2 * 1024 * 1024, rssBytes: 3 * 1024 * 1024 },
     });
+    // Exhaustive on purpose: this is the shape vega's shell parses, so an
+    // accidental addition should have to be acknowledged here.
     expect(payload).toEqual({
       status: 'ok',
       app: 'openswarm',
@@ -47,6 +50,9 @@ describe('buildHealthPayload', () => {
       backend_pid: 42,
       backend_parent_pid: 1,
       uptime_s: 12,
+      heap_used_mb: 1,
+      heap_limit_mb: 2,
+      rss_mb: 3,
     });
   });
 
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of a3305ff (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the health payload implementation, its production caller in src/support/web.ts:468, and the focused unit/live-server coverage. The metrics correctly use V8 heap statistics plus process RSS, convert bytes to rounded MiB, and are included in the unauthenticated health response.

PR #452, opened 2026-08-29
feat(coordination): assign agents handles that read like handles (AGT-4064)

unohee:
Closes AGT-4064.

## What the operator saw

```
[오전 11:51:25] Atlas 3 2 (worker · AX-1030) → reviewer-b0bc: Codename: Atlas 3
```

Two banned shapes in one line: `reviewer-b0bc`, the deterministic fallback, reads as a machine ID where a participant's name belongs; `Atlas 3 2` is a self-chosen name that has collected collision counters.

## Why the numbering escalated

The collision set was built from **every entry in a 500-slot registry** — every name registered since boot, including agents that finished hours ago:

```ts
const takenAddresses = new Set([...chosenAgentNames.values()].map((v) => v.address));
for (let n = 2; takenAddresses.has(address) || RESERVED_FALLBACK_ADDRESS.test(address); n += 1) {
  candidate = `${cleaned} ${n}`;
```

So "Atlas" collided with a **dead** agent and became "Atlas 2", then "Atlas 3". Those bumped names reached later agents through the board history they read, so a model reported `Codename: Atlas 3` as its own choice and was bumped again. **The numbering fed itself.**

## What changed

Agents no longer name themselves. Identity is assigned from role-flavoured vocabulary in four shapes:

| shape | worker | reviewer |
|---|---|---|
| Word + RoleWord + 4 digits | `NimbusDev1892` | `WardenQA7061` |
| word `_` roletag | `copper_dev` | `caliper_rv` |
| Word + role suffix | `FurrowBuilds` | `AmberNotes` |
| word + word + 2 digits | `harborsolder19` | `indigosentry44` |

Over 1600 generated identities: **0** role-hex, **0** numeric suffixes, **0** doubled words, all four shapes present, 79% distinct. A collision advances to a **different** handle, never a decorated one.

**Identity is deliberately not scoped by session.** A retry opens a fresh session, and if that changed the handle then an answer addressed to the first attempt would sit in an inbox nobody reads. An existing test pinned that invariant and my first attempt broke it — the session-keyed registry gave every retry a new handle. Caught by that test, not by me.

**The handle is stamped over the model's `codename`** where the stage result is produced. Linear reads that field through `trackerEffects.completionStats`, so without the stamp the board would read `NimbusDev1892` while Linear still said `Atlas 3`.

## Verification

Mutations, each killing exactly the intended tests:

| mutation | tests killed |
|---|---|
| restore the role-hex generator | banned shape; role vocabularies; board addressing; review-agent naming |
| append a collision counter instead of a new handle | collision resolves to a different handle |
| emit one shape only | handle shape varies |
| share one vocabulary across roles | reviewer vs worker vocabulary |
| drop the codename stamp | assigned handle reaches `codename` |

The stamp test drives the **real** pipeline (`new PairPipeline(...).run()` with `runWorker` mocked). My first version of it stamped the value itself inside the test and so passed with the production change reverted — rewritten.

Four tests pinning self-naming were removed rather than adapted: they covered behaviour that no longer exists. One test in `periodicReview.test.ts` asserted the **banned** `role-hex` shape and now asserts its absence.

`npm run typecheck` clean, `oxlint` clean, **113 tests green** across the eight files touching the identity surface. Commit gate: `openswarm review` ×2 — APPROVE, then APPROVE again after the retry-stability fix.

## Incidental

`repoNameFromPath` / `worktreeNameFromPath` move to `src/agents/repoPathNames.ts` to keep `pairPipeline.ts` under the 1500-line pre-commit cap. They are **also copied** in `src/tui/subagentTree.ts` and `src/automation/runnerExecution.ts` — consolidating all three is worth doing but is not this change's job, and the new module says so.

## Renaming agents that are already running

Identity lives in an in-memory map, so the redeploy carrying this reassigns every agent, including tasks already in flight. Correlation is by `correlationId`, not by name, so a mid-run rename does not break an exchange — a reader will see one participant's name change once.

### DIFF
--- a/src/agents/pairPipeline.board.test.ts
+++ b/src/agents/pairPipeline.board.test.ts
@@ -5,8 +5,7 @@ vi.mock('../coordination/runCoordination.js', () => ({
   publishCoordination: vi.fn(async (event: Record<string, unknown>) => { published.events.push(event); }),
 }));
 
-const { publishStageToBoard, publishStageOutcomeToBoard, registerChosenAgentName, stageCorrelationId } = await import('./pipelineCoordination.js');
-const { assignCallSign } = await import('../coordination/agentNames.js');
+const { publishStageToBoard, publishStageOutcomeToBoard, assignedAgentName, stageCorrelationId } = await import('./pipelineCoordination.js');
 
 function context(over: Record<string, unknown> = {}) {
   return {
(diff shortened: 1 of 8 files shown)


unohee:
## 🔍 Fresh review of 7ce1208 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The assigned-handle wiring correctly replaces model codenames in pipeline results and board events, but collision-resolved identities are process-local. A daemon restart (or cache eviction) can change an active participant’s address, invalidating replies addressed to the prior handle; persist the resolved identity or make collision resolution restart-stable.

unohee:
## 🔍 Fresh review of ccf3f2b (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked deterministic handle assignment, cross-role collision avoidance, board/MCP/Linear identity wiring, periodic-review call-sign wiring, and the extracted path helpers. The implementation consistently replaces model-provided codenames before persistence and uses task-scoped mailbox consumption, so no material defect was found.

PR #453, opened 2026-08-29
feat(coordination): let an agent wait for a reply instead of being killed for waiting (AGT-4065)

unohee:
Closes AGT-4065. First step toward the operator's goal: **conversation and replies that flow in real time**.

## The finding

There is **no wait primitive anywhere in the system**. Every path writes to the board, returns immediately, and hopes the addressee polls before its run ends. `CoordinationStore.consume` is pure mark-and-return (`coordinationStore.ts:420-432`) — an empty inbox comes back instantly.

And an agent that tried to wait by polling **was killed for it**:

```ts
// agenticLoop.ts:555-560
if (allToolCallsSeen(toolCalls, seenToolCalls)) {
  noProgressTurns++;
  if (noProgressTurns >= NO_PROGRESS_LIMIT) { /* = 3 */ break; }
```

`allToolCallsSeen` keys on `name + args`, and `coordination_read` takes **no parameters** — so every check produced an identical key. Three checks of a quiet inbox looked exactly like a stalled model.

## What changed

**1. Checking an inbox is not a stall.** Tools whose answer changes without this agent doing anything are exempt from the repetition judgement. The loop stays bounded by `maxTurns` and its wall clock, so this cannot produce an unbounded run.

**2. `coordination_wait(timeout_ms)`** — returns as soon as something addressed to this agent lands, or an empty list at the deadline. Clamped hard (default 20s, max 60s): a wait sits inside a stage with its own budget (worker 20 min, reviewer 6 — `stageTimeouts.ts`) and must never become a way to burn it and fail as `infra_error`.

## Three defects the commit gate found in my own implementation

| round | finding | fix |
|---|---|---|
| 1 | **subscription race** — draining before subscribing left a window where a message reached neither path, and the agent waited out its whole deadline on mail that had arrived | subscribe first, then drain |
| 1 | **in-process only** — the hub cannot see `openswarm attach` or the CLI, which write to the board file from their own processes. The wait would silently always time out for exactly the operator path it exists to serve | a 2s re-drain alongside the hub |
| 3 | **an unreadable board looked like a quiet one** — a swallowed store failure returned `[]`, which an agent reads as "nobody answered" and proceeds on | report it as an error; a transient failure that later succeeds still resolves normally |

And one I introduced while fixing the first: guarding against overlapping drains made a wake-up arriving **mid-drain** get dropped, so the message that woke us would wait for the next re-drain — or forever. It is now remembered and re-checked.

## Incidental

The adapter kept its **own hardcoded copy** of the coordination tool names (`tools.ts:840`), so a newly declared tool could be defined, advertised to the model, and still fall through to "Unknown tool". It now derives them from the definitions, and a test drives every declared name through the real dispatch.

## Verification

Seven mutations, each killing exactly the intended tests:

| mutation | tests killed |
|---|---|
| stall exemption removed | 3 inbox-check tests |
| no initial drain | pre-arrived mail |
| never subscribes to the hub | early return on arrival |
| clamp removed | timeout clamping |
| adapter reverts to a hardcoded list | every-declared-tool dispatch |
| no cross-process re-drain | wake for an unseen publisher |
| mid-drain wake-up dropped | mid-drain wake-up |
| store failure swallowed | unreadable board reported |

The mid-drain race is milliseconds wide against the real store, so that test drives `waitForInbox` with a fake whose first drain is still running when the event fires — a timing-based test could not hit it reliably. `waitForInbox` is exported for that reason, and the comment says so.

`npm run typecheck` clean, `oxlint` clean, **130 tests green** across the six affected files. Commit gate: `openswarm review` ×4 — REVISE ×3 (above), then APPROVE.

## Deliberately not in this change

- Splitting `ask_human` from "stop this run" — welded in three places; it deserves its own change now that waiting exists.
- Relaxing `consume`'s `taskId` filter, which makes cross-task peer messaging write-only.
- Worker/reviewer overlap. Judge it after these land.

### DIFF
--- a/src/adapters/agenticLoop.test.ts
+++ b/src/adapters/agenticLoop.test.ts
@@ -639,3 +639,41 @@ describe('compactPriorTurns with a wide tool fan-out', () => {
     expect(fanOutHasNoOrphans(messages)).toBe(true);
   });
 });
+
+// `allToolCallsSeen` keys on name+args, and `coordination_read` takes no
+// parameters — so every inbox check produced an identical key and three checks
+// of a quiet inbox tripped the stall detector. An agent waiting for a reply was
+// killed for waiting. (AGT-4065)
+describe('checking an inbox is not a stall', () => {
+  const call = (name: string, args = '{}') => ({ id: `c-${name}-${args}`, function: { name, arguments: args } });
+
+  it('does not count a repeated inbox check as a stalled turn', async () => {
+    const { allToolCallsSeen } = await import('./agenticLoop.js');
+    const seen = new Set<string>();
+    const turn = [call('coordination_read')];
+    // Simulate three consecutive identical checks, as an agent awaiting a reply
+    // would make.
+    for (let i = 0; i < 3; i += 1) {
+      expect(allToolCallsSeen(turn, seen)).toBe(false);
+      seen.add('coordination_read:{}');
+    }
+  });
+
+  it('does not count a wait as a stalled turn either', async () => {
+    const { allToolCallsSeen } = await import('./agenticLoop.js');
+    expect(allToolCallsSeen([call('coordination_wait', '{"timeout_ms":5000}')],
+      new Set(['coordination_wait:{"timeout_ms":5000}']))).toBe(false);
+  });
+
+  it('still catches a genuine stall on tools that read only this agent\'s own work', async () => {
+    const { allToolCallsSeen } = await import('./agenticLoop.js');
+    const repeated = [call('read_file', '{"path":"a.ts"}')];
+    expect(allToolCallsSeen(repeated, new Set(['read_file:{"path":"a.ts"}']))).toBe(true);
+  });
+
+  it('a turn that also checked the inbox is not a stall, even when its other calls repeat', async () => {
+    const { allToolCallsSeen } = await import('./agenticLoop.js');
+    const mixed = [call('read_file', '{"path":"a.ts"}'), call('coordination_read')];
+    expect(allToolCallsSeen(mixed, new Set(['read_file:{"path":"a.ts"}', 'coordination_read:{}']))).toBe(false);
+  });
+});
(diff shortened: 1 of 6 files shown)


unohee:
## 🔍 Fresh review of aebf065 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The coordination tool registration/dispatch is now consistently derived, and the wait implementation covers initial drain, in-process wakeups, cross-process polling, and read failures. However, the wait timers are unref'd while the agentic loop is awaiting them, which can let a CLI process exit before the requested wait completes when these are its only active handles.

unohee:
## 🔍 Fresh review of 42c750f (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The dispatch-name derivation and wait wake-up/error paths are wired and covered. However, the new wait timeout is only capped to a fixed 60 seconds, not to the active agent-loop/stage time remaining, so it can overrun the loop deadline it claims to respect.

unohee:
## 🔍 Fresh review of afa80a0 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the current diff and production wiring. `coordination_wait` is advertised and dispatched from the shared definition-derived name set, its loop deadline is propagated and clamped, and the wait implementation handles event-hub wakeups, cross-process polling, read failures, and destructive in-flight drains; focused tests cover these paths.

unohee:
## 🔍 Fresh review of afa80a0 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Reviewed the current coordination wait, loop-deadline propagation, dispatcher wiring, and associated tests. The wait correctly distinguishes a board that was never readable, but it still reports an empty inbox when the board becomes unreadable after its initial empty drain.

unohee:
## 🔍 Fresh review of 0a6c90f (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the full changed coordination wait implementation and its production wiring. `coordination_wait` is advertised, dispatched from the shared definition set, receives the agent-loop deadline, safely handles destructive inbox consumption and cross-process publishing, and is covered for timeout, wake-up, read-failure, and deadline-race paths.

PR #454, opened 2026-08-29
feat(orchestration): read the graph as conversations about tasks (AGT-4066)

unohee:
Closes AGT-4066.

The orchestration view drew every identity it had ever seen as a floating dot in one `EXECUTION` band, so six live tasks read as forty agents. This makes the graph read as a set of conversations instead.

## What changed

**Task lanes.** The execution tier is now one lane per task, holding that task's own participants; operator, control plane and coordination stay as shared rails above. `node.taskId` already carried the grouping and was simply never drawn.

A node that has only ever been *addressed* now joins the task it was addressed within, so a reviewer who never speaks first still sits with its worker. A node that has *acted* keeps its own task, so cross-task advice still cannot drag a working agent out of its lane — the existing invariant is unchanged and still tested.

**Bounded by liveness, not history.** The client kept every event the server ring ever handed it, so a dashboard left open for a shift held more history than the daemon does. `byId` now holds at most the ring's own 2000 events (a ceiling checked after every absorb, pruned in batches so the sort is amortised). Agents outside the 30-minute window drop out of the graph behind a `show idle (N)` toggle. A scheduled sweep wakes at the moment the next agent ages out, so a silent board still sheds them rather than freezing whatever was last drawn.

**Stable positions.** `x` was `(index + 0.5) / rowMembers.length`, so one arrival re-indexed and slid everyone. A node's column now comes from a hash of its address, and nodes are seated oldest-first — so an arrival can only take a free slot and never displaces an incumbent, even when it hashes onto one. Lane order is pinned from the *unfiltered* board, so hiding an idle agent cannot re-stack the lanes either (the vertical version of the same defect).

**No more teardown.** `svg.innerHTML = ''` on every SSE event is replaced by keyed enter/update/exit over persistent layers, so the rare real move can transition instead of jumping. One delegated-lifetime listener per node group, so redraws cannot stack listeners.

**Interaction floor.** An "N agents are waiting on you" banner above the graph (`stats` already had the data, buried among six chips); search, task and role filters, and the idle collapse. Edge tooltips name agents by handle rather than by routing address (AGT-4064).

## Verification

`npx tsc --noEmit` clean, `oxlint web/ src/` clean (2 warnings, both pre-existing on `main` in files this PR does not touch), 210 web tests pass.

Every new test was mutation-checked: the production change it covers was reverted and the test confirmed to turn red. 27/27 mutations killed, including reverting `x` to the index formula, seating by name instead of age, removing the eviction, restoring `svg.innerHTML = ''`, dropping the lane-order pin, and reintroducing a raw NUL byte into the served asset.

Two defects were found by the commit gate and fixed here rather than deferred: the batch eviction let steady state sit above its stated 2000 bound, and the liveness window was never re-evaluated on a quiet board. A third — the resize listener outliving `stop()` — predates this branch but is fixed since it is in a file this PR owns.

## Deliberately not done

- **Zoom/pan.** There is no viewBox transform today; it is a larger change than it looks and the issue marks it optional.
- **`src/support/dashboardHtml.ts`.** Its inline ES5 duplicates some of this derivation. Out of scope per the issue, and verified independent — it imports none of these modules.
- **Composer draft preservation.** A redraw rebuilds the thread panel and loses unsent text. Pre-existing (any SSE event already does this) and entangled with tested send/accept semantics, so out of scope here.

### DIFF
--- a/tests/web/orchestrationModel.test.ts
+++ b/tests/web/orchestrationModel.test.ts
@@ -7,7 +7,14 @@
 
 import { describe, expect, it } from 'vitest';
 // @ts-expect-error — browser ESM asset without type declarations
-import { buildOrchestrationModel, dominantKind, KIND_COLORS } from '../../web/static/js/orchestrationModel.mjs';
+import {
+  buildOrchestrationModel, dominantKind, filterGraphNodes, KIND_COLORS, taskLanesOf,
+} from '../../web/static/js/orchestrationModel.mjs';
+
+type Node = {
+  id: string; name: string; role: string; taskId?: string; taskLabel?: string;
+  active: boolean; firstSeen: number; lastSeen: number;
+};
 
 let seq = 0;
 function event(over: Record<string, unknown> = {}) {
@@ -76,7 +83,7 @@ describe('buildOrchestrationModel', () => {
     expect(roles['adapter-router']).toBe('daemon');
   });
 
-  it('tracks the task a node last ACTED in, ignoring tasks it merely received from', () => {
+  it('prefers the task a node ACTED in over any task it was merely addressed within', () => {
     const model = buildOrchestrationModel([
       event({ taskId: 'task-1' }),
       // Cross-task advice TO this node must not drag it into task-9.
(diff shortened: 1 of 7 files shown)


unohee:
## 🔍 Fresh review of 358ce8f (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Reviewed the current model/view/layout wiring and lifecycle paths. The stopped-view cleanup is incomplete: an already-started refresh can still mutate and redraw after stop(), recreating DOM/timers and retaining the view state.

unohee:
## 🔍 Fresh review of e78dd65 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the current model, view, layout, HTML wiring, and focused web tests. The bounded event retention, liveness expiry/cleanup, stable lane ordering, filter controls, keyed SVG reconciliation, and text-safe edge-key implementation are wired into production callers and covered by behavior tests; no material defect found.

PR #455, opened 2026-08-29
fix(worktree): stop a container restart wedging the admission cap for 24h (AGT-4067)

unohee:
Closes AGT-4067.

## The wedge

vela, 2026-08-29, 80 minutes after a redeploy: **6 `NEEDS_RECONCILE` rows holding all 6 admission slots**, 24 `READY` tasks unable to be admitted, reconciler logging `Keeping … in NEEDS_RECONCILE (active_owner)` 67× per row.

The marker on disk for AX-858:

```json
{ "ownerPid": 7, "createdAt": "2026-08-29T01:21:00.946Z" }
```

Written at 10:21 KST by the *previous* container. The new daemon was handed **pid 7 again** — container pid assignment is deterministic — so `processAppearsAlive(7)` returned true. The only remaining guard was the 24-hour `activeMarkerAbandonMs` window (AGT-4053), and the marker was 4 hours old.

AGT-4056 built the exit path out of `NEEDS_RECONCILE`. It was deployed and running. It never fired, because the evidence it consults said the dead owner was alive.

## The fix

A pid is unique among live processes. A marker recording **this process's own pid** was written either by us or by a process that has since exited; if its `createdAt` predates our start it cannot be ours, so its writer is provably gone. No new field required — which is why it also releases the markers already on disk.

`ownerInstanceId` is written alongside and read in **one direction only**: a *matching* id positively identifies a marker as ours. A *differing* id proves nothing (`openswarm attach`, a manual `openswarm run`, a second daemon are all "not us"), and treating it as proof would put two owners on one worktree.

Applied at all three liveness sites: `inspectWorktreeRecovery`, the `pruneWorktrees` orphan sweep, and `removePreservedWorktreeAt`'s re-claim guard — the last of which the daemon calls itself on every run conclusion, so a stale marker there silently leaked preserved worktrees.

Shortening `OPENSWARM_ACTIVE_MARKER_ABANDON_MS` was considered and rejected: it is a blunt age cutoff applied to every marker including a live 20-minute stage's, so it takes worktrees away from running executors.

## Tests — 5 mutations, each verified to fail

| Mutation | Test that dies |
|---|---|
| predates-proof removed | all 3 "predates this process" tests (recovery / prune / removePreserved) |
| differing id treated as dead (the first draft's bug) | "respects a marker from another live OpenSwarm process" |
| any id treated as dead | "keeps a marker from this boot for the full window" + 2 pre-existing |
| legacy (no id) treated as dead | "falls back to the pid + age test" + 1 pre-existing |
| third site's boot check removed | "cleans up when the only active marker predates this process" |

78/78 in `worktreeManager.test.ts` + `worktreeManager.coverage.test.ts`; 98/98 across the reconciler-adjacent suites (`autonomousRunner.durable`, `runLedger`, `durableRunCoordinator`). `tsc` clean.

## Gate

Commit gate `openswarm review` (tier 1): **REVISE** round 1 — the first draft read a *differing* `ownerInstanceId` as proof of death, which would let any concurrent OpenSwarm process take a live worktree. Fixed as described. **APPROVE** round 2.

## Remaining verification

The DoD's last item is deliberately not yet checked: the six wedged rows must leave `NEEDS_RECONCILE` **without a manual DB write** after deploy. They are being left in place on purpose — letting the fix clear them is the test.

### DIFF
--- a/src/support/worktreeManager.coverage.test.ts
+++ b/src/support/worktreeManager.coverage.test.ts
@@ -286,6 +286,84 @@ describe('createWorktree retry/resume error paths', () => {
     }
   });
 
+  // The regression this fix must not introduce: `openswarm attach`, a manual
+  // `openswarm run` and a second daemon are all "not this process", so a
+  // differing owner id cannot be read as proof of death. Doing so would put
+  // two owners on one worktree. (Caught by the commit-gate review.)
+  it('respects a marker from another live OpenSwarm process even though its owner id differs (AGT-4067)', async () => {
+    const info = await createWorktree(repo, 'INT-1', 'swarm/INT-1-sibling-process');
+    const markerPath = join(
+      repo, '.git', 'openswarm', 'active-worktrees', 'INT-1', `${info.activeMarkerToken}.json`,
+    );
+    const marker = JSON.parse(readFileSync(markerPath, 'utf8'));
+    // A different process that is genuinely running: our own parent. Its id
+    // differs from ours and its marker predates nothing — it must be kept.
+    marker.ownerInstanceId = 'sibling-process-00000000-0000-4000-8000-000000000000';
+    marker.ownerPid = process.ppid;
+    marker.createdAt = new Date(Date.now() - 4 * 60 * 60 * 1000).toISOString();
+    writeFileSync(markerPath, JSON.stringify(marker));
+
+    await expect(inspectWorktreeRecovery(repo, 'INT-1', info.worktreePath)).resolves.toMatchObject({
+      state: 'active_owner',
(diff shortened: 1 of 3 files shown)


unohee:
## 🔍 Fresh review of be88c21 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the marker write/read/recovery, terminal cleanup, pruning, and durable-reconciler call paths. The new pre-start/self-PID proof is applied consistently where PID reuse previously blocked recovery, while differing instance IDs correctly remain insufficient to take a concurrently owned worktree; integration tests cover recovery, pruning, and terminal cleanup.

PR #456, opened 2026-08-29
fix(liveness): stop a stale task-state lock costing 10 minutes per restart (AGT-4068)

unohee:
Closes AGT-4068.

## The ten minutes

Deploying AGT-4067 to vela at 15:12:43 KST bought ten minutes of this:

```
[HB] ✗ Linear fetch error: Timed out waiting for task state lock: …/task-state.json.lock
[AutonomousRunner] Task fetch failed (3x consecutive)
```

The lock, frozen throughout:

```json
{"pid":7,"token":"b0c68f3a-…"}      mtime 15:12:42.710  ← 0.66 s BEFORE the recreate
```

Written by the *previous* container's daemon, also pid 7. The new daemon is pid 7 too, so `isProcessAlive(7)` answered "alive" about its own ghost. It recovered at ~15:25 when the age rule fired — and only then could the AGT-4067 reconciler release the six wedged runs. Every restart pays this.

**Fourth call site of the same trap** (AGT-4023, AGT-4052/4053, AGT-4067), each fixed in place with its own local `isProcessAlive` copy. The proof now lives in one `src/support/processLiveness.ts`.

## Namespace, because that is the scope a pid is unique in

`store.test.ts` already pinned that a lock inside the window must be respected because "it may belong to a live writer in another pid namespace sharing the mounted state file". Both the lock and the worktree marker are mounted state two containers can share, each with its own pid 1. So a record naming another space is not reasoned about by pid **at all** — not the proof, and not the liveness probe, whose `ESRCH` would otherwise reclaim a live owner's resource. It falls back to the age rule, or to "keep" at `removePreservedWorktreeAt`, which has none.

Three states, kept distinct: a **string** identifies the space; **`null`** records that the writer could not determine its own; **absent** means the record predates the field and keeps the original local probe. `null` is written explicitly because an omitted key is indistinguishable from the third case.

## Gate — 6 real defects over 7 rounds

| # | Finding |
|---|---|
| 1 | the shared helper regressed both call sites' fail-closed handling — a blanket `catch` made `EPERM` (a live process owned by someone else) and an unusable pid read as *dead* |
| 2 | worktree markers applied the same-pid proof with no namespace recorded |
| 3 | a Linux process that cannot read `/proc/self/ns/pid` collapsed to a host-wide default, so two such containers would compare equal |
| 4 | `removePreservedWorktreeAt` still probed foreign-namespace pids locally; a local `ESRCH` authorised deleting a tree a live container owns |
| 5 | the symmetric hole in the lock: `abandonedLock` probed before the namespace check |
| 6 | "unknown namespace" serialized as an *omitted* key, indistinguishable from a legacy record |

Four of those were holes in the namespace safety I was adding — the reviewer tracked the invariant more consistently than I did. Round 7: **APPROVE**.

## Mutation evidence

11 mutations, each verified to fail. Two initially **survived** and the tests were rewritten, not kept:

- `namespacesMatch(undefined, undefined)` and the Linux-collapse fallback are both unreachable on macOS, so a mutation to either passed the whole suite. Extracted `resolveNamespaceId(platform, readLink)` and `namespacesMatch(recorded, mine)` as pure functions and tested those directly — all three branches now die under mutation on any platform.
- A mutation removing the redundant `=== null` guards survived because `null` can never equal a string id anyway. Replaced with the mutation that describes the actual defect — treating `null` as legacy — which kills both new tests.

184 tests across `processLiveness`, `store`, `worktreeManager` ×2, `autonomousRunner.durable`, `runLedger`. `tsc` clean.

## Verification still outstanding

The DoD's last item needs a restart to observe: the first heartbeat after a deploy should fetch tasks instead of timing out. Will be checked on the next vela deploy.

### DIFF
--- a/src/support/processLiveness.test.ts
+++ b/src/support/processLiveness.test.ts
@@ -99,23 +99,58 @@ describe('processNamespaceId / sameProcessNamespace', () => {
     expect(namespacesMatch('ns-a', 'ns-b')).toBe(false);
   });
 
-  // Both branches below are unreachable on macOS, so they are exercised through
+  // The branches below are unreachable on macOS, so they are exercised through
   // the pure resolver rather than the platform. Without this, a mutation that
-  // collapses an unreadable Linux namespace to a shared host-wide default
-  // passes the whole suite. (AGT-4068)
+  // collapses an unreadable Linux namespace to a shared default passes the
+  // whole suite. (AGT-4068)
+  const BOOT = 'boot-1c3bda3d-e21a-4166-be21-c34ce6538059';
+  const NS = 'pid:[4026531836]';
+
   it('reports an unreadable Linux pid namespace as unknown, not as a shared default', () => {
-    const id = resolveNamespaceId('linux', () => { throw new Error('EACCES /proc/self/ns/pid'); });
+    const id = resolveNamespaceId('linux', () => { throw new Error('EACCES /proc/self/ns/pid'); }, () => BOOT);
     expect(id).toBeUndefined();
   });
 
-  it('identifies a readable Linux pid namespace by host and namespace together', () => {
-    const id = resolveNamespaceId('linux', () => 'pid:[4026531836]');
-    expect(id).toBe(`${hostname()}:pid:[4026531836]`);
+  it('reports an unreadable boot id as unknown too', () => {
+    const id = resolveNamespaceId('linux', () => NS, () => { throw new Error('EACCES boot_id'); });
+    expect(id).toBeUndefined();
   });
 
-  it('uses the host name alone where the platform has no pid namespaces', () => {
-    const read = vi.fn(() => 'pid:[4026531836]');
-    expect(resolveNamespaceId('darwin', read)).toBe(hostname());
-    expect(read).not.toHaveBeenCalled();
+  it('separates two hosts that share a namespace inode', () => {
+    // Inodes are a per-boot counter, so two machines hand out the same numbers
+    // routinely. Without the boot id these would compare equal and each could
+    // reclaim the other's live lock across a shared state directory.
+    const a = resolveNamespaceId('linux', () => NS, () => 'boot-aaaa');
+    const b = resolveNamespaceId('linux', () => NS, () => 'boot-bbbb');
+    expect(a).not.toBe(b);
+    expect(namespacesMatch(a, b)).toBe(false);
+  });
+
+  it('separates two containers on one host, which share a boot id', () => {
+    const a = resolveNamespaceId('linux', () => 'pid:[4026531111]', () => BOOT);
+    const b = resolveNamespaceId('linux', () => 'pid:[4026532222]', () => BOOT);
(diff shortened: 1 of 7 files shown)


unohee:
## 🔍 Fresh review of 9906ea4 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The new liveness wiring is production-reachable and correctly distinguishes absent versus explicitly unknown pid namespaces. However, its namespace identity is not safe across hosts: hostname plus a Linux pid-namespace inode can collide, allowing a process on another host sharing the state directory to reclaim a live owner’s lock/worktree.

unohee:
## 🔍 Fresh review of 8e9bf91 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The Linux namespace proof is wired into both lock and worktree reclamation and has targeted coverage. However, the non-Linux identity falls back to hostname, which is not a unique machine or PID-space identity for shared state; this can authorize immediate reclamation of a live remote owner.

unohee:
## Note on the last commit-gate finding — premise verified false, no code change

The final gate round said the `pidns:` prefix is *"incompatible with namespace strings already persisted by the prior release, turning same-space records into foreign/unjudgeable records after upgrade."*

No such records exist. Checked rather than assumed:

```
$ git show origin/main:src/support/worktreeManager.ts | grep -c ownerNamespace   → 0
$ git show origin/main:src/taskState/store.ts         | grep -c "ns:"            → 0
$ docker exec openswarm grep -c ownerNamespace /app/dist/support/worktreeManager.js → 0
$ docker exec openswarm grep -c pidns /app/dist/support/processLiveness.js         → MISSING (no such file)
$ git log --oneline origin/main..HEAD  → both AGT-4068 commits, neither merged
```

No released version writes a namespace field at all, and the running deployment has no `processLiveness` module. The intermediate formats (`<hostname>:<ns>`, then `<bootId>:<ns>`) existed only on unmerged commits of this branch and were never deployed.

Records written by the released code have the field **absent**, which is a distinct third state from `null` and keeps the original local probe unchanged — now pinned by `still probes a marker written before the pid-space field existed (AGT-4068)`, which fails when `absent` is folded into `null`.

Earlier rounds in this PR were all acted on; this one is the exception, with the evidence above.

unohee:
## 🔍 Fresh review of 113e018 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The Linux pid-namespace proof is wired into both lock and worktree recovery and its same-namespace/predating-process tests are present. However, non-Linux `host:` identifiers are still treated as sufficient to probe and reclaim a matching-host record, despite hostname equality not proving a shared PID space.

unohee:
## 🔍 Fresh review of 113e018 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The Linux pid-namespace proof is correctly wired into both marker and task-state-lock reclamation, and its fail-closed unknown-namespace handling is covered. However, matching non-Linux hostname hints still authorize a local PID probe, which can reclaim a live remote owner on a shared state directory when hosts share a hostname.

PR #457, opened 2026-08-29
fix(discord): let a reply to an agent's question answer it (AGT-4070)

unohee:
Closes AGT-4070. Reported by the operator:

> 디스코드에 답장하기 기능으로 답변 보냈는데 오류: `<urlopen error [Errno 61] Connection refused>` 라고 뜨는데 전달이 안되는거아냐?

The `urlopen` error is **not** OpenSwarm's — this bot is TypeScript, and in six hours it logged exactly one inbound chat message on an unrelated subject. vela also runs a separate Python VEGA stack (`vega-all`, `vega_router` on :18000), which is where that came from.

What *is* ours is that the reply would not have answered anything even had it arrived.

## The dead end

```ts
// src/discord/discordCore.ts — before
if (ALLOWED_USER_IDS.includes(msg.author.id) && !msg.content.startsWith('!')) {
  await handleChat(msg);   // every reply landed here
  return;
}
```

The only path to `answerHumanQuestion` was `!answer <correlation-id> <text>`, with an id copied by hand out of the message. **28 runs are parked** on `[operator-question] asked 2 times with no answer`, against **116 unanswered questions** on the board.

## The link lives in Discord, not in this process

A reply carries `message.reference.messageId`. Rather than keeping a map from message id to correlation id, the correlation id is read back out of the `!answer` hint on the message being replied to. That matters because a parked question routinely outlives the daemon — asked twice, parked, answered hours later, possibly after a restart — and an in-memory map would have lost exactly the ones that matter.

Only our own messages are read for it: otherwise an operator could reply to any message containing that text, including one they wrote themselves, and answer through a route that never passed the allowed-user check on the question.

## Gate — 2 real defects over 3 rounds

| # | Finding |
|---|---|
| 1 | correlation extraction scanned the whole message, and the **agent-authored question body comes first** — a question that merely discusses `!answer <id>` would hand the reply the wrong id |
| 2 | a long question is split across messages and the hint is appended last, so **the chunk that shows the question was a dead end to reply to** |

(1) moved the format into `src/coordination/answerHint.ts`, which both sides import, and anchored the match to a whole line with the last one winning. (2) added `chunkForDiscord`, which repeats the hint on every chunk. Round 3: **APPROVE**.

## Mutation evidence — 8 mutations, each verified to fail

| Mutation | Test that dies |
|---|---|
| drop the bot-author check | "ignores a message somebody else wrote, however it is worded" |
| drop the `!botUserId` guard | "claims nothing when … we do not know who we are" — `undefined !== undefined` is false, so an author-less message would read as ours |
| match anything, not the hint | "reads the correlation id out of a question we posted" + 1 |
| unanchored `!answer` scan | "finds nothing in a message that carries no hint" |
| first hint line wins | "prefers the generated line over an earlier one the body reproduces in full" |
| prefix without the closing anchor | "rejects a line that opens like the hint but does not end like it" |
| hint on the last chunk only | "repeats the hint on every chunk of a split question" |

Two of these tests were rewritten after their first version let the mutation survive: comparing chunkers at different `maxLen`, and cases where line order and the closing anchor happened not to matter.

40 tests across `answerHint`, `discordCore.helpers`, `discordCore.lifecycle`, `discordDelivery`, `humanQuestions`. `tsc` clean.

## Not changed

`!answer <id> <text>` works exactly as before; the posted instruction now offers both. Access control is untouched — only `DISCORD_ALLOWED_USERS` can answer either way.

### DIFF
--- a/src/coordination/answerHint.test.ts
+++ b/src/coordination/answerHint.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from 'vitest';
+import { answerHint, correlationIdFromHint } from './answerHint.js';
+
+// AGT-4070: a posted question is the agent's own words followed by a generated
+// instruction, and a reply is matched back by reading that instruction. The
+// agent's text comes FIRST, so an unanchored scan would let a question that
+// merely discusses `!answer` hand the reply the wrong correlation id.
+// (Caught by the commit-gate review.)
+describe('answerHint / correlationIdFromHint', () => {
+  it('round-trips the correlation id it wrote', () => {
+    expect(correlationIdFromHint(answerHint('hq-abc123'))).toBe('hq-abc123');
+  });
+
+  it('reads the generated line, not a correlation id mentioned in the question body', () => {
+    const posted = [
+      'OpenSwarm needs a decision for AX-1 (asked by kestrel_build).',
+      'The runbook says to use !answer hq-WRONG when a job stalls — should we keep that?',
+      '',
+      answerHint('hq-RIGHT'),
+    ].join('\n');
+    expect(correlationIdFromHint(posted)).toBe('hq-RIGHT');
+  });
+
+  it('ignores a hint-shaped line the body only quotes in passing', () => {
+    // A prefix match is not enough: the whole line has to be the hint.
+    const posted = `See the note: ${answerHint('hq-QUOTED')} (do not do this yet)\n\n${answerHint('hq-REAL')}`;
+    expect(correlationIdFromHint(posted)).toBe('hq-REAL');
+  });
+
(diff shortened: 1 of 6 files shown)


unohee:
## 🔍 Fresh review of 88aecb1 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked production wiring from `postHumanQuestion` through Discord delivery and reply routing, including author verification, anchored/last-hint parsing, restart-safe correlation lookup, and chunk sizing with repeated hints. The changed tests cover the previously ambiguous body-hint and split-message cases; no material defect found.

PR #458, opened 2026-08-29
fix(liveness): decide a prior generation by owner id, not by clock (AGT-4071)

unohee:
Closes AGT-4071.

## AGT-4068 shipped and did not fire

Deployed 17:12 KST. First heartbeat after the restart:

```
[HB] ✗ Linear fetch error: Timed out waiting for task state lock
```

— the exact symptom AGT-4068 closed. The measurement:

```
lock mtime         17:12:12.670   ← written by the OUTGOING container at shutdown
container started  17:12:13.231
gap                0.561 s        vs PROCESS_START_JITTER_MS = 1.000 s
```

Everything else was working: the pid space matched and was proof-capable on both sides (`pidns:1c3bda3d-…:pid:[4026533598]` — the namespace inode is stable across recreates, which is worth knowing). Only the margin was wrong, and it was wrong in the one case that exists.

**Why the AGT-4068 verification missed it.** The synthetic lock planted to isolate the fast path was backdated *five minutes*, which sails past a 1s margin. The parameter did not represent the real distribution.

## Shrinking the margin makes it worse

The margin absorbs an imprecise timestamp, and for the lock that timestamp is the file's **mtime**. On a filesystem with one-second granularity a lock this process wrote at `…13.900` reports `…13.000`, before our own start of `…13.231` — and we would reclaim our own live lock. The margin has to be larger than the gap it must resolve, so the timestamp path cannot decide this case at all.

## Decide by identity

A pid is unique among live processes. A record carrying **our pid** whose **owner id is not ours** was written by a process that has exited. No clock, no granularity, no NTP step.

This is not the "a differing id means dead" rule AGT-4068 rejected — that failed because a *live sibling* process also differs. The pid conjunct is what excludes it, and it is load-bearing: dropping it kills two tests.

The worktree marker already recorded `ownerInstanceId` and only needed reading. The task-state lock now records one too. The timestamp comparison remains the fallback for records carrying no owner id — which is every record written before this.

## Mutation evidence — 5 mutations, each verified to fail

| Mutation | Test that dies |
|---|---|
| remove the owner-id path | "proves death from the owner id even when the record is newer than the clock margin" + 1 |
| a matching owner id also counts as gone | "never reclaims our own record, however far back its timestamp reads" |
| drop the pid conjunct | "claims nothing about a pid that is not ours" + "still claims nothing about another live process" |
| lock stops recording `instance` | "records who holds it, so a successor can decide ownership without a clock" |
| (earlier) namespace and fail-closed gates | unchanged from AGT-4068, still covered |

That fourth one initially **survived**: `withStoreLock` removes the lock file in its `finally`, so the writer's output cannot be observed after the fact. Extracted `buildLockPayload` so the tests can see it.

The lock integration tests now use the **production timing** — a record 0.5 s before our start, inside the margin — rather than a comfortably backdated one.

139 tests pass, 6 skipped (the Linux-only pid-space cases, which run in CI). `tsc` clean.

## Gate

Commit gate: **APPROVE** on the first round. Its own vitest run was terminated at the time limit (exit 143); the 139/6 figure above is from running the same suites directly.

## DoD not yet met

The last item needs the next deploy: verified against a **real container recreate**, not a backdated synthetic record — the first heartbeat must fetch with no lock timeout. That is precisely the check AGT-4068 did not do.

### DIFF
--- a/src/support/processLiveness.test.ts
+++ b/src/support/processLiveness.test.ts
@@ -72,6 +72,45 @@ describe('writerProvablyGone', () => {
   it('claims nothing when the timestamp is unreadable', () => {
     expect(writerProvablyGone({ pid: process.pid, writtenAtMs: Number.NaN })).toBe(false);
   });
+
+  // AGT-4071: the timestamp path could not decide the case it was written for.
+  // A container recreate takes about half a second — measured on vela at 0.561 s
+  // — which is INSIDE any margin big enough to absorb a coarse mtime. The owner
+  // id settles it without consulting a clock at all.
+  it('proves death from the owner id even when the record is newer than the clock margin', () => {
+    expect(writerProvablyGone({
+      pid: process.pid,
+      ownerId: 'a-previous-boot',
+      ourOwnerId: 'this-boot',
+      writtenAtMs: PROCESS_STARTED_AT_MS - 500, // the real gap; the margin is 1000
+    })).toBe(true);
+  });
+
+  it('never reclaims our own record, however far back its timestamp reads', () => {
+    // The margin existed for this: a lock file's mtime can round down below our
+    // own start on a coarse filesystem. With an owner id the timestamp is not
+    // consulted, so it cannot go wrong.
+    expect(writerProvablyGone({
+      pid: process.pid,
+      ownerId: 'this-boot',
+      ourOwnerId: 'this-boot',
+      writtenAtMs: 0,
+    })).toBe(false);
+  });
+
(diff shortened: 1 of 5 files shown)


unohee:
## 🔍 Fresh review of 37372ad (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the current production wiring for both task-state locks and active-worktree markers. The new instance ID is persisted at write time, is used only after same proof-capable PID-namespace gating, and falls back to the timestamp path for legacy records; tests cover both the restart and coarse-mtime cases.

PR #459, opened 2026-08-29
fix(adapters): let the worker's report reach the reviewer (AGT-4073)

unohee:
## The defect

The worker prompt (`src/locale/prompts/{ko,en}.ts:230-233`) asks for a plain-text
report whose **first line** is the agent's `Codename:` introduction, and says no
JSON is needed on the success path. `extractSummary()` kept only `lines[0]`, so:

- `src/agents/reviewer.ts:215-219` built the reviewer prompt's
  `- **Summary:** Codename: Atlas` — the worker's words never reached the reviewer.
- `src/agents/pipelineCoordination.ts:158-162` stripped the codename, got an empty
  string, and the coordination board fell back to `Finished in 148.8s`.

## Measured on the deployed daemon (6 h of `coordination_trace`)

| role | timing fallback | real speech |
|---|---|---|
| reviewer | 2/9 (22 %) | **78 %** |
| worker | 38/61 (62 %) | 38 % |

Reviewers post substantive `[revise]` critiques; workers answered with a duration.
That asymmetry was a missing channel, not a difference in model output.

## Proven by running the parser, not by reading it

Feeding `parseWorkerResult()` exactly the shape the prompt asks for:

```
before: summary = "Codename: Atlas"          → board: timing fallback
after:  summary = "Wired a low-confidence filter into renderA2() … still unverified."
```

## The change

- Skip `Codename:` lines before choosing the summary.
- Join whole lines up to the **existing** 200-character bound, so a report's later
  sentences (what could not be verified, where to look) survive. The bound is
  unchanged — this is not a widening.
- `extractSummary` existed byte-for-byte in `resultParsing.ts`, `codex.ts` and
  `claude.ts`; it now lives once and the other two import it.

## Verification

- 31 tests in `resultParsing.test.ts` pass; 6 are new.
- Three mutations verified to fail a test: removing the codename filter, removing
  the line join, and removing the cap boundary. The cap test was rewritten after
  the first version failed to discriminate — length alone passes either way, so it
  now asserts the summary ends at a whole line rather than mid-word.
- Downstream consumers re-run green: `worker.test.ts`, `reviewer.test.ts`,
  `linear/format.test.ts`, `documenter.test.ts`, `codex.test.ts`, `claude.test.ts`
  (183 tests).

## Known limit, deliberately not changed here

An English three-line report needs ~240 characters, so the 200-character cap still
drops its last line. The cap is pre-existing and applied equally before this fix;
sizing a new one needs evidence about real report lengths, which is not collected
anywhere today. Recorded on AGT-4073 rather than widened unreviewed.

Closes AGT-4073.

### DIFF
--- a/src/adapters/claude.ts
+++ b/src/adapters/claude.ts
@@ -19,6 +19,7 @@ import type { ReviewDecision } from '../agents/agentPair.js';
 import { extractCostFromStreamJson, formatCost } from '../support/costTracker.js';
 import { extractResultFromStreamJson } from '../agents/cliStreamParser.js';
 import { t } from '../locale/index.js';
+import { extractSummary } from './resultParsing.js';
 import { writeClaudeMcpConfig } from './memoryMcp.js';
 
 const execFileAsync = promisify(execFile);
@@ -267,12 +268,6 @@ function extractWorkerFromText(text: string): WorkerResult {
   };
 }
 
-function extractSummary(text: string): string {
-  const lines = text.split('\n').filter((l) => l.trim().length > 10);
-  if (lines.length === 0) return t('common.fallback.noSummary');
-  const summary = lines[0].trim();
-  return summary.length > 200 ? summary.slice(0, 200) + '...' : summary;
-}
 
 function extractErrorMessage(text: string): string {
   const errorMatch = text.match(/(?:error|exception|failed?):\s*(.+)/i);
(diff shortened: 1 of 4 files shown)


unohee:
## 🔍 Fresh review of a8a7113 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the shared extraction implementation and its Claude/Codex production callers. The codename line is excluded, subsequent complete report lines are retained within the existing 200-character limit, and overlong first content lines retain prior truncation behavior; focused tests cover these paths.

PR #461, opened 2026-08-29
fix(runner): publish a parked run's committed work as a draft PR (AGT-4076)

unohee:
> supersded할거면 PR은 올리고 가라그래.

## The defect

A run that stops for an operator decision leaves its commits on a branch that was
**never pushed**. Measured on the deployed daemon:

```
swarm/AX-1030-ops-tool-solapi   9 commits ahead   remote: absent
swarm/AX-885-a1-3               6                 absent
swarm/AX-863-a2-4               3                 absent
swarm/AX-1066-backend-…-fix     2                 absent
swarm/AX-871-ext-a2-slack       2                 absent
swarm/AX-879-a2-9-2             1                 absent
```

23 commits, no PR, while the operator was being asked **70 questions** about work
they could not see.

## Where it publishes, and why not anywhere else

At the moment the run parks, in the runner, right beside the approved path. There the
executor still owns the claim, the lease and the worktree, so there is nothing to race.

An earlier attempt swept parked rows from the reconciler. The commit gate rejected it
three times for the same reason: a parked run can be resumed and claimed while an
out-of-band publish is in flight, and if that publish opened a PR first, the eventual
reviewed publication would silently reuse it and ship approved work as a draft.

## The rules this had to learn (each one a gate finding)

- **Lease fence.** Uses the same `beforePublish()` the approved path does, so an
  executor that already lost its claim cannot publish for work it no longer speaks for.
- **`committedOnly`.** Skips every write into the working tree — no marker strip, no
  `add -A`, no commit — so the resume continues from exactly where the worker stopped.
- **Draft.** Nothing reviewed this.
- **Promotion on reuse.** When the branch later passes review, both reuse paths (the
  pre-create lookup and the create-race fallback) take the PR out of draft, or reviewed
  work ships hidden.
- **…except when duplicates justify the draft.** A PR opened draft because another
  branch also closes the issue (INT-2544) stays draft. The duplicate set is re-checked
  rather than assumed.
- **Never re-ready an already-open PR.** `gh pr ready` fails on one, and that failure
  would mark a successful run as `infra_error`.
- **No `result.prUrl`.** Any prUrl on a non-approved result is classified as
  `publication_reconcile` and parked in NEEDS_RECONCILE — which would convert an
  operator park into a row holding a repository admission slot. That is the exact
  phantom-row shape that idled the whole loop earlier the same day. The PR is attached
  to the ledger instead.

## Verification

- 154 tests across `worktreeManager`, `publishOnPark`, `durableRunCoordinator` and
  `runnerExecution.coverage` pass.
- New tests use real git fixtures with a fake `gh` on PATH: committed work is pushed
  while a dirty file stays dirty; a reused draft is promoted; a reused non-draft is not
  touched; a duplicate keeps it draft; the create-race loser promotes.
- Seven mutations each fail a test: dropping `committedOnly`, dropping `draft`, dropping
  the promotion, promoting on the draft path, dropping the `isDraft` check, dropping the
  duplicate re-check, dropping the race-path promotion.
- A guard test in `durableRunCoordinator.test.ts` pins the classification that makes
  `result.prUrl` unsafe here, so the reasoning goes stale loudly rather than silently.

## Two extractions

Both forced by the 1500-line pre-commit cap, both coherent on their own:
`publishOnPark.ts` holds the two publication paths (same helper, differing only in what
the run earned) and `pullRequestReady.ts` holds the promotion rule shared by the two
PR-reuse paths.

Closes AGT-4076.

### DIFF
--- a/src/automation/durableRunCoordinator.test.ts
+++ b/src/automation/durableRunCoordinator.test.ts
@@ -303,6 +303,31 @@ describe('DurableRunCoordinator', () => {
     coordinator.close();
   });
 
+  it('a parked result carrying a prUrl stops being a park (AGT-4076 guard)', async () => {
+    // Why the parked-publish path in runnerExecution attaches the PR to the
+    // LEDGER but never writes `result.prUrl`. `publishedNeedsReconcile` treats
+    // any prUrl on a non-approved result as publication debt, so setting it
+    // would turn an operator park — which frees the repository admission slot
+    // and resumes on the answer — into a NEEDS_RECONCILE row that holds one.
+    // If this classification ever changes, the comment in runnerExecution.ts
+    // is stale and this test is where that shows up.
+    const coordinator = new DurableRunCoordinator({
+      mode: 'primary', dbPath: dbPath(), instanceId: 'park-with-pr',
+    });
+
+    await coordinator.execute(task('AGT-PARK-PR'), '/repo', async () => ({
+      ...result(false),
+      finalStatus: 'waiting_on_operator',
+      prUrl: 'https://github.com/o/r/pull/1',
+    }));
+
+    expect(coordinator.getRun('AGT-PARK-PR')).toMatchObject({
+      state: 'NEEDS_RECONCILE',
+      lastErrorCode: 'publication_reconcile',
+    });
+    coordinator.close();
+  });
+
   it('reconciles a published PR before any retry when publication attachment throws', async () => {
     const path = dbPath();
     const ledger = new RunLedger(path);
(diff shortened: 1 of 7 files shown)


unohee:
## 🔍 Fresh review of d250571 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the parked-run path through `runnerExecution`, durable publication attachment, and PR reuse/promotion. Parked work is published draft-only without mutating the pipeline result, preserving operator readmission semantics; approved reuse promotes the draft unless a duplicate issue PR still requires draft status.

PR #462, opened 2026-08-29
feat(agents): make an agent diagnose before it pages a human (AGT-4081)

unohee:
> 자격증명 워크트리에 넣고 (평소 CGF 작업을 클로드 코드에 맡겼던 것처럼 만들어줘야지.
> cxt도 같이 도구로 줘야할거 같아. 그러면 빠르게 스캔할거 아니야
> 1,2,3 다 구현해봐 cxt 사용 훅 포함해서

## The credentials were already there

Verified on vela before changing anything:

```
local-only paths linked into worktrees   6 / 6 paths,  24 / 24 worktrees
  .env, apps/pipelines/.env, apps/portal/.env,
  apps/portal/.dev.vars, infra/n8n/.env, docs/CGF_data
```

Readable from inside a worktree — key **names** only: `NAS_TAILSCALE_IP`,
`NAS_USERNAME`, `NAS_PASSWORD`, `ECOUNT_*`, `META_*`, `OPENROUTER_API_KEY`,
`JETTE_*`, 7 `SLACK*` keys, `WAREHOUSE_DSN`. AGT-4061's sandbox exception, which
lets the agent's file tools follow those symlinks, is deployed.

Reachability probed from the container with the real values (never printed):

```
NAS over Tailscale, port 22    reachable
Warehouse PostgreSQL           ConnectionRefused
```

Yet **40 of 74** unanswered operator questions asked for exactly this material.

## What was actually missing

```
docker   OK        psql        MISSING
curl     OK        pg_isready  MISSING
python3  OK        ssh         MISSING
node     OK        jq          MISSING
gh       OK        nc          MISSING
```

The parked questions blame a missing `docker`. **`docker` is installed.** An agent
that cannot finish a check describes the first plausible cause it can think of, and
that guess reaches the operator as fact.

## Three changes, one cause

1. **Toolchain** — `postgresql-client`, `openssh-client`, `jq`, `netcat-openbsd`.
   No socket mount, no new credential, no privilege change: clients only, for
   services the container already reaches.

2. **Diagnose before escalating** — "do not guess" applied to implementation but not
   to diagnosis, so one failed attempt became "impossible". The worker prompt and the
   `ask_human` tool description now require three checks with their output pasted into
   the question: the path (`ls -la`, `readlink -f` for symlinks), the tool
   (`command -v`), and the credential (**key names only**, never a value).

3. **cxt** — installed, so the prompt no longer forbids `cxt scan`. An unfamiliar
   repository gets indexed once instead of read file by file, and
   `who-calls`/`impact` answer "what breaks if I change this".

## A real drift the gate caught

The prompt documented `impact`/`who-calls`/`export` while `package.json` still pinned
`@intrect/cxt@^0.1.0`, which has none of them — and `buildWorkerEnv` **prepends** the
bundled `.bin`, so 0.1.0 is what agents would have got. Telling an agent to run a
command that does not exist is the very failure this PR exists to stop.

Pinned to `0.3.0` (matching the Dockerfile) and added `cxtCommandSet.test.ts`, which
extracts every `` `cxt <verb>` `` the prompt names and asserts the bundled CLI
advertises it. Reverting to 0.1.0 fails it with
`prompt documents \`cxt who-calls\` but the bundled CLI has no such command`.

## Verification

- 80 tests across `prompts`, `cxtCommandSet` and `coordinationTools` pass.
- `@intrect/cxt@0.3.0` verified to install and run on the image's own base
  (`node:22-bookworm-slim`).
- Three mutations each fail a test: removing the diagnosis requirement (en), restoring
  the `cxt scan` prohibition (ko), and downgrading the bundled CLI to 0.1.0.

Closes AGT-4081.


unohee:
## 🔍 Fresh review of 6ad76cc (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the worker PATH wiring in src/adapters/base.ts:157 and envPath.ts:25-51: the local bundled cxt now resolves from the lockfile at 0.3.0, matching the Docker global fallback and documented commands. The new prompt/test coverage is wired to both locales; the historical local-vs-global cxt version mismatch is resolved.

PR #463, opened 2026-08-29
feat(observability): catch event-loop stalls from inside the daemon (AGT-4079)

unohee:
## Why

External probing established that the daemon blocks its event loop for tens of seconds under dispatch load, and logs **nothing** while it happens.

Measured on vela with a 200-sample `/api/health` probe (that handler is pure synchronous property reads ahead of both auth gates, so its latency *is* loop lag):

| bucket | samples |
|---|---|
| < 0.5 s | 196 |
| 0.5–2 s | 1 |
| 2–10 s | 1 |
| **≥ 10 s** | **2** |

The two large ones: **28.58 s**, and one that returned `30.00` — the probe's own `--max-time`, so read it as ≥30 s, not as a measurement. The second sat inside a **43-second gap with no daemon output at all**.

Seven candidate mechanisms were eliminated from outside: the `/api/stats` handler, knowledge-graph JSON (74 ms parse on a real 20 MB graph), knowledge scans (9 and 19 minutes from the stalls), embedding inference (354 ms per call), the preflight PR-overlap query (1.8 KB, 0 ms parse), the lifecycle lock, the board parse. Every survivor lives inside that silent gap.

## What this adds

`src/support/eventLoopMonitor.ts` — an always-on watcher armed at daemon start. When a window's worst delay crosses the threshold it emits one line naming the delay and, from CPU accounting, **whether the loop was computing or parked**. Those two have disjoint fixes, and picking wrong costs a full investigate-and-deploy cycle.

```
[EventLoop] Blocked 999ms (mean 250ms over 1200ms) — cpu-bound: on-CPU 83.8% of the window in user code (user 991ms) — synchronous JS [fs r0/w0 ctx v0/i979]
[EventLoop] Blocked 1014ms (mean 254ms over 1216ms) — blocked-off-cpu: off-CPU 99.8% of the window (user 2ms, sys 1ms) — parked waiting, not computing [fs r0/w0 ctx v0/i16]
```

## Two things measurement changed mid-implementation

**`perf_hooks.monitorEventLoopDelay` does not work for this.** It is the obvious instrument, and a smoke test against a real block killed it: on Node v26 an 800 ms busy-wait left its `max` at the 21 ms warm-up value, and a 900 ms `Atomics.wait` left it at **0**. Replaced with timer lateness — the same approach that had already caught the 354 ms block a single embedding call causes.

**The classifier is narrower than first written.** An earlier version also branched on `fsRead`/`fsWrite` and voluntary-vs-involuntary context switches. Measured against real stalls those counters do not support it: macOS `getrusage` leaves fs counts and voluntary switches at zero even across 12 `readFileSync` calls, and *involuntary* switches track being pre-empted **while busy** (830 during a 1 s CPU burn vs 20 during a 1 s blocking wait) — the opposite of the "starved by other tenants" reading they had been given. They now ride along in the log line as raw evidence and decide nothing.

## Verification

- 12 tests; **6 mutations killed** (unit conversions, the kernel-vs-JS split, the fs gate, the short-window guard, the reporter containment).
- End-to-end against real stalls, both arms, shown above.
- A re-arm test pins that a retry after a failed `start()` replaces the watcher rather than double-reporting.
- `tsc --noEmit` clean, oxlint clean.

## Lifecycle

Armed before `engine.init()` so a startup stall is caught too (the embedding model load alone blocks ~780 ms). `start()`'s body is split into `startAfterMonitor()` — body unchanged, no reindent — so a rejection has exactly one path that releases the watcher; `unref()` only frees the process to exit, it does not stop the interval.

Review gate: tier 1 (`openswarm review`, codex gpt-5.6-terra), 3 rounds — 2 findings, both on the failed-start lifecycle, both fixed. `Decision: APPROVE` on round 3.

Closes part of AGT-4079: this does not name the blocking frame, it settles which class it belongs to. The next measurement is the deploy.

### DIFF
--- a/src/automation/autonomousRunner.ts
+++ b/src/automation/autonomousRunner.ts
@@ -56,6 +56,7 @@ import {
   removePreservedWorktreeAt,
 } from '../support/worktreeManager.js';
 import { loadRepoMetadata } from '../support/repoMetadata.js';
+import { startEventLoopMonitor } from '../support/eventLoopMonitor.js';
 import { STUCK_LABEL } from '../linear/index.js';
 import { refreshGraph, toProjectSlug } from '../knowledge/index.js';
 import { checkAllMonitors, getActiveMonitors } from './longRunningMonitor.js';
@@ -190,6 +191,7 @@ export class AutonomousRunner {
   private _heartbeatRunning = false;
   private heartbeatCompletion: Promise<void> | null = null;
   private stopPromise: Promise<void> | null = null;
+  private stopEventLoopMonitor: (() => void) | null = null;
   private deferredShutdownCleanup: Promise<void> | null = null;
   private durableRunsClosed = false;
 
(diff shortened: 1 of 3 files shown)


unohee:
## 🔍 Fresh review of e12c053 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the production lifecycle wiring: the monitor is armed before startup work, is cleaned up on startup rejection and stop, and has a real caller in `AutonomousRunner` (`src/automation/autonomousRunner.ts:1480-1491,1621-1624`). The timer-lateness sampling, resource-usage delta conversion, reporter containment, and threshold/classification coverage are internally consistent (`src/support/eventLoopMonitor.ts:155-244`; `src/support/eventLoopMonitor.test.ts:27-169`).

PR #465, opened 2026-08-29
perf(task-state): stop rewriting the whole store for unchanged issues (AGT-4086)

unohee:
## What

Every heartbeat, `service.ts:231` reconciles **every** issue fetched from Linear (~780) via
`updateTaskLinearState`, which persisted unconditionally. The store is one file holding all
tasks, so a single unchanged row still cost a full read + `JSON.parse` + zod validate +
`JSON.stringify(store, null, 2)` + write + `fsync` + rename + directory `fsync`.

## Measured on vela (2026-08-29)

| quantity | value |
|---|---|
| tasks in store | 3,146 |
| file size | 1,882,801 B |
| `JSON.parse` | 9.6 ms |
| `JSON.stringify(…, null, 2)` | 8.5 ms |
| **per unchanged row** | **18.1 ms + 2 fsyncs** |

Effect: the daemon's event loop was blocked **~30 s out of every ~60 s**, writing **1.39 GiB
per sweep** (27 GB in the half hour after a restart). Per-thread attribution during a stall put
2.92 s of a 3 s window on the main thread with 149 MB written, while four V8 GC helper threads
took ~10% each — which is the reported on-CPU 106–129%.

The arithmetic closes both ways: 1.3937 GiB ÷ 1.883 MB = 777 rewrites, and the store reports
**784 rows with `updatedAt` inside the sweep window**. 718 of those 784 were `backlog` tasks
that had not moved since the previous poll.

## How

Splits the decision out as a pure `planLinearStateReconciliation(current, linearState)` that
returns `null` when the merged state would be identical; `updateTaskLinearState` then returns
the current row without writing.

**Freshness is unchanged.** The old code already computed its patch from the same *unlocked*
`getTaskState` read, and `ensureStoreLoaded` revalidates against `mtimeMs:size` on every call.
The one new window is another writer landing between our `stat` and the decision, which costs a
single missed reconciliation that the next heartbeat corrects.

## Tests

8 new cases, and the behavioural one asserts on the file's **mtime** rather than a return value,
so it fails if the write comes back. Mutation-tested:

| mutation | result |
|---|---|
| never skip (revert the fix) | **3 tests fail** |
| drop the same-status guard | **1 test fails** |
| skip even for an unknown task | **1 test fails** |
| `current &&` → `current?.` | survives — equivalent under the type contract (`linearState` is always a string) |

Full file: 34 passed / 3 skipped. Typecheck clean.

## Follow-ups (not needed to remove the stall)

- Batch the sweep so one lock + one persist covers N patches.
- Drop the `null, 2` pretty-printing (roughly halves both size and serialize time).

Closes AGT-4086. Found by the in-daemon monitor from AGT-4079 (#463).

### DIFF
--- a/src/taskState/store.test.ts
+++ b/src/taskState/store.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
-import { existsSync, mkdtempSync, readFileSync, rmSync, unlinkSync, utimesSync, writeFileSync } from 'node:fs';
+import { existsSync, mkdtempSync, readFileSync, rmSync, statSync, unlinkSync, utimesSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { hostname, tmpdir } from 'node:os';
 import { spawn } from 'node:child_process';
@@ -17,6 +17,7 @@ import {
   buildTaskStateSyncComment,
   hydrateTaskStateFromComments,
   markTaskBacklog,
+  planLinearStateReconciliation,
   resetTaskStateStoreForTests,
   buildLockPayload,
   type OpenSwarmTaskState,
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 73920dc (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the production caller in `src/core/service.ts:226-247`, reconciliation/write path in `src/taskState/store.ts:448-510`, and task-state behavior/tests. The null-plan fast path avoids persistence only when both recorded Linear state and reconcilable execution state already agree; changed and unmapped states still persist, while stale live/resolved statuses retain the prior reconciliation behavior.

PR #467, opened 2026-08-29
feat(agents): tell a worker what its sibling worktrees are editing (AGT-4088)

unohee:
## Why

A worker talks only to its reviewer. It cannot learn what other workers on the **same
repository** are editing, so two can rewrite the same file and only find out when their
branches integrate.

The board cannot carry this: **50 of the 55** distinct `repository` values on the live
board are worktree paths, and reads filter on exact equality
(`coordinationStore.ts:260`). Same-repo conflict checking is also off on purpose —
`autonomousRunner.ts:1191`, *"must not serialize workers"*. The gap is not that overlap
is allowed; it is that **the worker is never told**.

## Current exposure

| measurement | value |
|---|---|
| open cgf-portal PRs | 5, all MERGEABLE |
| active tasks declaring a `fileScope` | 37 |
| active tasks declaring **none** | **104** |
| `maxConcurrentTasks` | 9 |

Zero conflicts today — but concurrency is running at 1–3, not 9, and `fileScope` is
empty for 74% of active tasks, so it could not be the basis for detection.

## How

Collect at dispatch what every other worktree has uncommitted; render as an advisory
block in both locales. Read-only — the worker is told to narrow its change, not to wait.

**git, not the run ledger.** git knows every worktree, including ones no run owns (the
main checkout, one left by an abandoned run), and those hold conflicting edits too; it
is also reachable from inside a worktree, which the ledger is not. Files come from the
working tree, not `fileScope`.

## Four defects the gate caught (5 rounds)

| # | defect | fix |
|---|---|---|
| 1 | worktree list capped **before** status was read — on this host (25 worktrees, cap 8) a dirty worktree past the cut vanished silently | inspect all, cap only the dirty ones |
| 2 | plain porcelain C-quotes unusual paths and prints renames as `old -> new`, so a filename containing `" -> "` split wrong and named a nonexistent file | `-z`, verified against **real git output** |
| 3 | per-command timeouts did not bound the scan: `ceil(N/concurrency)` batches ahead of *every* worker attempt | total budget |
| 4 | that budget skipped only *unstarted* worktrees, so running commands could overrun it | each call gets the remaining budget as its timeout |

Defect 1 was the one this feature exists to prevent — a silently omitted conflict.

Verification of the `-z` format against a real repository, not assumption:

```
A  a -> b.md\0        ← literal " -> " in a filename survives
 M my report.md\0     ← spaces literal, no quoting
R  new.ts\0old.ts\0   ← rename: new path, then source as its own field
```

## Tests

33 on the module, 4 on the rendered prompt. Every assertion verified to fail on revert,
including reintroducing each of the four defects above.

Closes AGT-4088.

### DIFF
--- a/src/agents/siblingWork.test.ts
+++ b/src/agents/siblingWork.test.ts
@@ -1,4 +1,7 @@
-import { describe, expect, it } from 'vitest';
+import { afterAll, describe, expect, it } from 'vitest';
+import { mkdtempSync, mkdirSync, rmSync, symlinkSync } from 'node:fs';
+import { join } from 'node:path';
+import { tmpdir } from 'node:os';
 import {
   MAX_FILES_PER_SIBLING, MAX_SIBLINGS, STATUS_CONCURRENCY, TOTAL_BUDGET_MS, collectSiblingWork,
   commandTimeoutMs, formatSiblingWork, identifierFromBranch, parseChangedFiles, parseWorktreeList,
@@ -84,6 +87,19 @@ describe('selectSiblingWorktrees', () => {
     // phantom peer.
     expect(selectSiblingWorktrees(entries, '/work/cgf-portal/worktree/../worktree/05a3c502')
       .map((e) => e.path)).not.toContain('/work/cgf-portal/worktree/05a3c502');
+  });
+
+  it('recognises ourselves when git reports our real path and we hold a symlinked one', () => {
+    // git always reports a worktree by its real path; the dispatched project
+    // path can carry a symlinked component. Compared lexically the two differ,
+    // and we would scan ourselves and report our own edits as a peer's.
+    const canonicalise = (p: string) => p.replace('/link/', '/real/');
+    const linked = [
+      { path: '/real/worktree/a' },
+      { path: '/real/worktree/b' },
+    ];
+    expect(selectSiblingWorktrees(linked, '/link/worktree/a', canonicalise).map((e) => e.path))
+      .toEqual(['/real/worktree/b']);
   });
 
   it('keeps the main checkout — it holds edits too', () => {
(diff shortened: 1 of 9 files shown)


unohee:
## 🔍 Fresh review of 63f4fc1 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Production wiring from pipeline context through both locale prompts is present, and status parsing correctly preserves unusual file paths. Fix worktree-list parsing: it corrupts valid whitespace-containing worktree paths, causing sibling changes to be silently missed or the current worktree to be reported as a sibling.

unohee:
## 🔍 Fresh review of 21a514c (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Production wiring from pipeline context through both locale prompts is present, and the status parser correctly uses NUL-delimited porcelain. Fix worktree-path decoding before merge: the parser treats porcelain path fields as literal despite Git C-quoting unusual paths, so it silently excludes valid siblings in such worktrees.

unohee:
## 🔍 Fresh review of e60393b (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The production wiring and prompt escaping are present, but self-exclusion is incorrect when the dispatched project path is a symlink. `resolve()` only normalizes lexical components and does not resolve symlinks, so the current worktree can be scanned and reported as a conflicting sibling.

unohee:
## 🔍 Fresh review of a1a4a3f (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked production wiring from pipeline context collection through worker prompt rendering, plus porcelain parsing, self-worktree exclusion, bounded status concurrency, and timeout handling. The implementation is reachable in worker execution and the added tests cover the substantive parser, symlink, cap, failure, and budget cases.

unohee:
## 🔍 Fresh review of 003de30 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked production wiring from `collectWorkerContext` through worker prompt rendering, bounded concurrent status collection, NUL porcelain parsing, path canonicalization, and EN/KO prompt isolation. The implementation handles the previously evident truncation, path-format, and self-symlink cases; no material defect found.

PR #468, opened 2026-08-29
fix(agents): keep OpenSwarm's own generated files out of the sibling advisory (AGT-4088)

unohee:
## Found by deploying it

AGT-4088 (#467) shipped and its own DoD check — run the collector against the live
repository — showed the advisory was mostly noise:

```
파일 항목 23개 중 16개가 .openswarm-preserved     (70%)
형제 19개 중 15개가 그 파일만 가짐                (79%)
```

Rendered, 6 of the 8 displayed lines were `AX-#### — .openswarm-preserved`, and the
four siblings doing real work were being pushed past the display cap by housekeeping.

Every synthetic fixture passed. Only the production distribution showed it.

## Fix

Filter the paths OpenSwarm generates inside a repository, and drop a sibling left with
nothing to report.

**An explicit list, not a `.openswarm/` prefix.** That directory also holds
`verify.yaml`, which is repository-owned configuration a task may genuinely change
(README:228; `deterministicTester.ts:13` `VERIFY_INPUTS`). Hiding a sibling's edit to it
would suppress exactly the overlap this advisory exists to report. Between over- and
under-filtering, over-filtering is the worse failure, so an unrecognised path is shown.

My first attempt *did* filter the whole directory; the commit gate caught it, and the
narrowing is covered by a test that fails if the prefix form comes back.

## Tests

46 on the module. Mutation-tested: disabling the filter, widening it to a prefix match,
and dropping it from the collector wiring each fail.

### DIFF
--- a/src/agents/siblingWork.test.ts
+++ b/src/agents/siblingWork.test.ts
@@ -5,7 +5,7 @@ import { tmpdir } from 'node:os';
 import {
   MAX_FILES_PER_SIBLING, MAX_SIBLINGS, STATUS_CONCURRENCY, TOTAL_BUDGET_MS, collectSiblingWork,
   commandTimeoutMs, formatSiblingWork, identifierFromBranch, parseChangedFiles, parseWorktreeList,
-  selectSiblingWorktrees,
+  selectSiblingWorktrees, withoutBookkeeping, isBookkeepingPath,
 } from './siblingWork.js';
 
 // `git worktree list --porcelain -z`: each attribute is its own NUL-terminated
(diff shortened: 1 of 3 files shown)


unohee:
## 🔍 Fresh review of 923fbb1 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the committed wiring: `collectSiblingWork` filters parsed porcelain status paths before dropping empty siblings (`src/agents/siblingWork.ts:127-139`), while the formatter remains import-safe for prompt modules. The explicit generated-path allowlist preserves repository-owned `.openswarm` configuration, and focused tests cover generated files, lookalikes, and collector behavior.

PR #469, opened 2026-08-30
fix(automation): reconcile a published run whose tracker card already went terminal (AGT-4094)

unohee:
## Problem

`reconcileDurableArtifacts()` skipped any `NEEDS_RECONCILE` row whose issue was absent from the heartbeat's task fetch:

```ts
if (!task) continue; // absence from an actionable fetch is ambiguous; keep parked
```

But the fetch **structurally excludes terminal cards** — `getMyIssues({ slim: true })` queries only `Todo`, `In Progress`, `In Review`, `Backlog` (`src/linear/linear.ts:670-680`). A run whose card reached Done therefore could never be seen by the reconciler again: skipped silently on every heartbeat, forever, while `claimRun()`'s admission query counts `NEEDS_RECONCILE` toward the per-project cap (`runLedger.ts:501-510`). The slot leaked permanently.

AGT-4056 deliberately narrowed its own fix to *branchless* orphans and routed a branch-carrying row to "`reconcileDurableArtifacts()`'s GitHub-by-branch lookup instead". That lookup never ran, because this guard short-circuited above it.

## Production evidence

**AX-876** (cgf-portal), run `f8c57098`: `pr_url` = merged PR #171, `owner_instance_id`/`lease_token` both NULL, card **Done at 12:46:21Z**, row entered `NEEDS_RECONCILE` at **13:15Z** — already terminal before it ever needed reconciling, so never once in a fetch. Stuck **11.9 h**, holding 1 of 9 slots. **Zero** occurrences of `AX-876` in the whole container log while `[Reconciler]` emitted 26 lines.

**Natural control — AX-862**: same project, same state, but a non-terminal card, so it stayed in the fetch → 26 `[Reconciler] Keeping AX-862 …` lines → `[Reconciler] Resuming worktree for AX-862`. Same state, opposite outcome, one differing variable.

## Change

Absence is genuinely ambiguous for the **worktree** path, which re-runs the work — that guard stays. It is *not* ambiguous for the **branch** path, where GitHub is the authority:

| PR state | Action | Safe without a live task? |
|---|---|---|
| merged / open | `recoverPublishedRun()` → DONE | yes — GitHub proves publication |
| closed unmerged | `markNeedsHuman()` | yes — a park, not a mutation |
| none | falls through to worktree recovery | **no** — still requires the task |

So the guard moves down to the one place it protects something, and the branch path rebuilds the task it needs from the durable row via a new pure helper `runRecordToTask()` — the exact inverse of `observeTask()`'s mapping. The completion effect reads only `issueId` / `issueIdentifier` / `title` / `linearProject`, all of which the record carries.

`reconcileDurableArtifacts` is only reached after a *successful*, non-empty fetch (`autonomousRunner.ts` returns early on `fetchResult.error` and `tasks.length === 0`), so "absent" here cannot mean "the fetch failed".

The remaining `!task` skip now logs, like every other branch of the loop — that silent exit is why this was invisible to log-reading in the first place.

## Verification

- `npx tsc --noEmit` clean; `npx vitest run src/automation/` — **636/636 pass**.
- **Revert-checked in both directions**, not just asserted:
  - guard restored to the top of the loop (pre-fix) → the new terminal-card test **fails**;
  - guard removed entirely (over-fix) → the parked-regression test **fails**;
  - as implemented → both pass. The AGT-4056 safety property is held by a test, not by a comment.
- `runRecordToTask` covered by a round-trip against `observeTask` (so the two mappings cannot drift), an unrecognized-source case (the column is a free-form string — it maps to `discovered` rather than being cast into the union), title fallbacks, and the no-project case.

## Concurrency

The branch path now runs for rows without a live task, so: the loop still returns early when `ownerInstanceId || leaseToken` is set, and `recoverPublishedRun` re-checks both **inside its transaction** plus `state_version` optimistic concurrency — a claim landing between the read and the write makes the update a no-op rather than a double-publish. The row's destination (`SYNC_PENDING` → `DONE`) is terminal, and `observeTask` reopens `DONE` only on `linearState === 'Todo'` or an explicit dispatch, so this cannot resurrect finished work. Effect delivery is idempotent via `dedupeKey` + marker comment.

Closes AGT-4094.

### DIFF
--- a/src/automation/autonomousRunner.durable.test.ts
+++ b/src/automation/autonomousRunner.durable.test.ts
@@ -384,6 +384,132 @@ esac
     internal.durableRuns.close();
   });
 
+  // AGT-4094: the heartbeat fetch asks Linear only for Todo / In Progress /
+  // In Review / Backlog, so a run whose card reached Done is structurally
+  // absent from `tasks`. Reconciliation used to skip such a row before it ever
+  // reached the GitHub lookup, leaving it in NEEDS_RECONCILE forever — and
+  // NEEDS_RECONCILE counts toward the per-project admission cap, so the slot
+  // leaked permanently. Production case: AX-876, merged PR, 11.9h stuck.
+  it('reconciles a published run whose tracker card already went terminal, so its admission slot is returned', async () => {
+    const [{ AutonomousRunner }, execution] = await Promise.all([
+      import('./autonomousRunner.js'),
+      import('./runnerExecution.js'),
+    ]);
+    const { updateProjectAfterTask } = await import('../linear/projectUpdater.js');
+    const bin = join(root, 'terminal-bin');
+    const repo = join(root, 'terminal-repo');
+    mkdirSync(bin, { recursive: true });
+    mkdirSync(repo, { recursive: true });
+    writeFileSync(join(bin, 'gh'), `#!/bin/sh
+case "$*" in
+  *"pr list --head"*) echo '[{"url":"https://github.com/acme/repo/pull/171","state":"MERGED","isDraft":false,"headRefOid":"abc171"}]';;
+esac
+`);
+    chmodSync(join(bin, 'gh'), 0o755);
+    const logPairComplete = vi.fn(async () => {});
+    execution.setTaskSource({
+      kind: 'linear',
+      fetchTasks: vi.fn(async () => []),
+      getExecutionComments: vi.fn(async () => []),
+      updateState: vi.fn(async () => true),
+      addComment: vi.fn(async () => {}),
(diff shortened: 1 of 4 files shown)


unohee:
## 🔍 Fresh review of 99a12be (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the reconciliation path and durable-task reconstruction. `src/automation/autonomousRunner.ts:1253-1290` now performs GitHub artifact lookup before treating an absent actionable task as a reason to park, and `src/automation/durableRunCoordinator.ts:149-168` rebuilds the completion-effect fields from persisted data. The added tests cover terminal-card published recovery, unpublished absence parking, source fallback, and persisted-field round-tripping.

PR #470, opened 2026-08-30
fix(automation): stop a parked task superseding itself via its own draft PR (AGT-4095)

unohee:
## Problem

`findOpenPRFileOverlaps()` has **no notion of which issue is being dispatched** — its signature is `(repoPath, plannedFiles)`. So when `applyDraftGates()` asks "does an open PR already own the files I plan to touch?", the answer includes **the task's own draft PR**, opened moments earlier by `publishOnPark` (AGT-4076) when the very same task parked on an operator question.

The task supersedes itself, on every heartbeat, forever.

## Production evidence (cgf-portal — the daemon's only enabled project)

Every supersede in the log matches the dispatched issue to a PR on **its own branch**:

| Superseded task | Blocking PR | PR branch |
|---|---|---|
| AX-858 | #180 | `swarm/`**`AX-858`**`-a3-4-sheet-lead-id` |
| AX-863 | #179 | `swarm/`**`AX-863`**`-a2-4` |
| AX-864 | #178 | `swarm/`**`AX-864`**`-a2-slack` |
| AX-1039 | #173 | `swarm/`**`AX-1039`**`-b1` |

4 of 4. Of the 8 open PRs on the repo, **7 are the daemon's own `swarm/*` drafts**.

**Throughput:** `automation_attempts` for `/work/cgf-portal` over the last 6 h holds `superseded` ×15 and `waiting_on_operator` ×4 — **zero real work attempts** — while each pointless dispatch still pays for a full Draft pre-analysis first (measured `[AX-874] [Draft] Complete in 56204ms`, one model call each).

**It ratchets.** `retryAtFor()` doubles the backoff on `superseded`. AX-863's own history: att5 (08-28 15:39Z) → att6 (19:12Z) → att13 → att15 → att16 (08-30 00:50Z), widening to the 6-hour ceiling. Six issues now sit 4.6–5.9 h out at attempt 10–16 without one genuine work failure among them.

Before AGT-4076 a parked run had no PR, so the gate had nothing to collide with. Neither change is wrong alone; the interaction is.

## Change

`findOpenPRFileOverlaps` takes an optional `selfIssueIdentifier` and drops PRs whose `headRefName` belongs to that issue; `applyDraftGates` passes `task.issueIdentifier ?? task.issueId`.

**Matched on the identifier, not the built branch name** — and this was computed against the live branches rather than reasoned about:

```
MATCH swarm/AX-858-a3-4-sheet-lead-id | PR: swarm/AX-858-a3-4-sheet-lead-id
MATCH swarm/AX-863-a2-4               | PR: swarm/AX-863-a2-4
DIFF  swarm/AX-864-a2-6-slack         | PR: swarm/AX-864-a2-slack
```

AX-864's title changed after its branch was cut, so an equality test against `buildBranchName` would have silently missed a third of real cases. The trailing delimiter keeps the prefix unambiguous: `swarm/AX-86-` cannot match `swarm/AX-863-…`.

Omitting the identifier keeps the pre-fix behaviour rather than disabling the gate.

## Refactor note

`worktreeManager.ts` sat one line under the 1500-line pre-commit cap, so the naming convention moved to its own `branchNaming.ts` — it is self-contained and now lives in one place. Consumers import it directly, which means two coverage suites that mocked `buildBranchName` *through* `worktreeManager` no longer intercept it. Those mock declarations are **removed rather than left inert**: the function is pure and deterministic, so the suites now exercise the real convention, and a stale mock would mislead the next reader.

## Verification

- `npx tsc --noEmit` clean; `npx vitest run src/automation/ src/support/ src/cli/workCommand` — **1271 passed, 3 skipped** (the skips are Linux-only pid-space cases).
- **Revert-checked:** deleting the self-exclusion line makes the new test fail; restored, it passes. The gate's actual purpose is held by an assertion in the same test — a *sibling* issue's overlapping PR still supersedes.
- `isBranchForIssue` covered for title drift (the AX-864 case), sibling-identifier ambiguity (`AX-86` must not claim `AX-863`), a bare slug-less branch, foreign branches, and an empty identifier.

Closes AGT-4095.

### DIFF
--- a/src/automation/draftGrooming.ts
+++ b/src/automation/draftGrooming.ts
@@ -47,7 +47,15 @@ export async function applyDraftGates(options: {
   }
 
   if (options.worktreeMode && draft.relevantFiles.length > 0) {
-    const overlaps = await findOpenPRFileOverlaps(options.projectPath, draft.relevantFiles);
+    // Exclude this issue's own PR. `publishOnPark` opens a draft PR for a run
+    // that parks on an operator question, so from the next heartbeat the task
+    // was colliding with itself and superseding forever — measured on
+    // cgf-portal as 15 supersedes and zero real work attempts in 6h. (AGT-4095)
+    const overlaps = await findOpenPRFileOverlaps(
+      options.projectPath,
+      draft.relevantFiles,
+      task.issueIdentifier ?? task.issueId,
+    );
     if (overlaps.length > 0) {
       const lines = overlaps.map(overlap => `- ${overlap.url}: ${overlap.files.map(file => `\`${file}\``).join(', ')}`);
       console.warn(`[AutonomousRunner] Existing open PR owns planned files — skipping duplicate worker: ${lines.join(' ')}`);
--- a/src/automation/runnerExecution.coverage.test.ts
+++ b/src/automation/runnerExecution.coverage.test.ts
@@ -40,7 +40,6 @@ const plannerNeedsDecomposition = vi.fn();
 const plannerEstimateTaskDuration = vi.fn();
 const plannerRunPlanner = vi.fn();
 const plannerFormatPlannerResult = vi.fn();
-const buildBranchName = vi.fn();
 const createWorktree = vi.fn();
 // Stand-in for the real class — the instanceof check under test resolves
 // through this same mock, so identity (not origin) is what matters (AGT-4038).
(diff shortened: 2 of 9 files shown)


unohee:
## 🔍 Fresh review of 4368070 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the production call chain from draft gating through PR overlap filtering, plus branch-name relocation callers and the real `gh`-stub overlap tests. The self-PR exclusion is delimiter-safe, preserves pre-existing behavior when no identifier is supplied, and handles title-slug drift without suppressing neighboring issue branches.

PR #472, opened 2026-08-30
fix(automation): reserve a PR's files only while a worker holds its run (AGT-4097)

unohee:
## Problem

`applyDraftGates` supersedes a task when an open PR already owns the files it plans to touch. That is right while someone is editing them — but `publishOnPark` leaves a PR behind long after its worker exits, so **an open PR alone is not evidence of anyone editing**. AGT-4095 removed the collision with the task's *own* PR; this removes it with every sibling's.

## Production evidence

Right after the AGT-4095 deploy, AX-864 drafted successfully (22 files) and the gate ran for real:

```
Existing open PR owns planned files — skipping duplicate worker:
  - .../pull/180: apps/pipelines/src/cgf_pipelines/jobs.py
  - .../pull/179: apps/pipelines/src/cgf_pipelines/a2_inputs.py
```

Its own #178 is correctly gone (AGT-4095 works), but #180 and #179 are PRs left behind by parked runs, untouched for 7–12 h. **Seven of the repo's eight open PRs are such leftovers**, and they have **zero pairwise file overlap** with each other — verified by computing the matrix from live `gh pr list --json files` — so this was never a deadlock between them. A broad plan simply sweeps in files that PRs going nowhere happen to own. AX-858/863/864 sat at attempts 14–16, 294–360 min out, each blocked attempt doubling the backoff again.

## The signal is the lease, not the reason a run stopped

My first implementation keyed on `lastErrorCode` (`NEEDS_HUMAN`, or `RETRY_AT` under `waiting_on_operator`). **It would not have fixed anything.** Checked against the live ledger before shipping:

```
AX-1028  RETRY_AT     superseded            176
AX-1029  RETRY_AT     superseded            174
AX-1030  NEEDS_HUMAN  needs_human           172
AX-1039  RETRY_AT     superseded            173
AX-854   RETRY_AT     superseded            170
AX-858   RETRY_AT     superseded            180
AX-859   RETRY_AT     waiting_on_operator   181
AX-863   RETRY_AT     superseded            179
AX-864   RETRY_AT     superseded            178
AX-879   RETRY_AT     superseded            177
→ predicate matched 2 of 10
```

Every one of those runs *had* parked on an operator question — that is why `publishOnPark` opened its PR — but each later dispatch overwrote `lastErrorCode`. A lease is not overwritten by the next attempt: it is held or it is not.

`isDraft` is not the signal either, checked rather than assumed: `worktreeManager.ts:1267` opens a PR as a draft for duplicates too, and says so in its own comment.

## Three distinctions the gate has to keep, each held by a test

| Case | Behaviour | Why |
|---|---|---|
| Branch outside `swarm/` | always reserves | no run to consult (a human's PR, another tool's) |
| `activeIssueIdentifiers: []` | nothing reserves | an assertion: the set of held leases is empty |
| `activeIssueIdentifiers: undefined` | every swarm PR reserves | the caller has no ledger to ask — `off` and `shadow` never claim, so they answer `undefined` rather than `[]` |
| Active row, lapsed lease | does not reserve | the process died; the row is waiting for a reconciliation sweep. Same expiry test the ledger itself uses |

Conflating the middle two would empty the gate on any caller that simply never wired the accessor — the mutation test for that line fails when the `??  []` shortcut is restored.

## Verification

- `npx tsc --noEmit` clean; `npx vitest run src/automation/ src/support/` — **1228 passed, 3 skipped** (Linux-only pid-space cases).
- Every new behaviour revert-checked: removing the lease guard, the `undefined`/`[]` distinction, or the self-exclusion each fails its own test.
- Commit gate ran **four rounds**. Rounds 1–3 each produced a real finding — the `lastErrorCode` predicate not reaching production state, `off`/`shadow` asserting an empty set, and lapsed leases still counted as held. Round 4: `Decision: APPROVE`.

## Note on scope

`worktreeManager.ts` tripped the 1500-line pre-commit cap again, one commit after being shaved to 1498. The pure overlap arithmetic (`computeFileOverlaps`, `formatOverlapReport`) moves to `fileOverlap.ts` — real headroom (now 1476) instead of another trimmed line. Re-exported, so no import path changes.

## Accepted tradeoff

Two branches may now edit one file. That is safe while they run (worktree isolation) and surfaces at integration, where `formatOverlapReport` already annotates the PR and AGT-4078 tracks merge conflicts on their own terms. The operator chose this explicitly over leaving siblings blocked indefinitely by one unanswered question.

Closes AGT-4097.

### DIFF
--- a/src/automation/autonomousRunner.ts
+++ b/src/automation/autonomousRunner.ts
@@ -2523,6 +2523,7 @@ export class AutonomousRunner {
       maxReflections: this.config.maxReflections,
       durability,
       peerIssues: this.lastFetchedTasks,
+      getActiveWorkerIssues: (p) => this.durableRuns.activeWorkerIdentifiers(p),
       mcpPolicies: this.config.mcpPolicies,
       adapterRouting: this.config.adapterRouting,
     };
--- a/src/automation/draftGrooming.ts
+++ b/src/automation/draftGrooming.ts
@@ -30,6 +30,8 @@ export async function applyDraftGates(options: {
   peers?: TaskItem[];
   source: ITaskSource | null;
   worktreeMode?: boolean;
+  /** Issues a worker currently holds; only their PRs reserve files. (AGT-4097) */
+  activeWorkerIssues?: readonly string[];
 }): Promise<PipelineResult | null> {
   const { task, draft, peers, source } = options;
   const duplicateTarget = peers?.find(peer => (peer.issueId || peer.id) === draft.duplicateOfIssueId);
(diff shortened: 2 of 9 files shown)


unohee:
## 🔍 Fresh review of 70856f4 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the primary-mode lease-to-draft-gate wiring, path normalization, self-PR exclusion, and swarm-vs-external branch filtering. The implementation preserves fail-closed behavior when lease state is unavailable and tests cover held, lapsed, empty, and unknown lease sets.

PR #473, opened 2026-08-30
fix(automation): honour the configured decomposition limits (AGT-4122)

unohee:
## Problem

Two independent defects let the planner create far more issues than the operator configured. Both fired on the first heartbeat after vega-agent was enabled and produced **23 Linear issues in two minutes** against a configured daily cap of 5.

```
[AutonomousRunner] Decomposition complete: 9 sub-issues created
[AutonomousRunner] Registered decomposition: parent=e414ecb5-…, children=9, daily=23/20
[AutonomousRunner] Daily issue creation limit reached: 23/20 — skipping decomposition
```

`daily=23/20` is decisive: the ceiling in force is the code default **20**, not the configured **5**.

**1. `service.ts` dropped most of the config.** It forwarded `enabled`, `thresholdMinutes`, `plannerModel` and `plannerTimeoutMs` and never set `decomposition` itself, so `maxDepth`, `maxChildrenPerTask`, `dailyLimit` and `autoBacklog` never reached the runner and every `??` default won. `AutonomousConfig.decomposition` already existed and the loader already parsed every field — only the hand-off was missing.

**2. `maxChildrenPerTask` never capped the planner.** Its only use was a pre-check against a task that *already* had children, so a plan of 9 created 9.

## Change

`service.ts` passes the whole object rather than a hand-picked subset, and the test asserts the object — so a field added later fails loudly instead of silently defaulting.

The cap now applies to the decomposition itself, and it **refuses rather than truncates**: taking the first N would leave the parent looking decomposed with scope the planner judged necessary silently dropped. Falling through to direct execution is what the daily-limit branch above already does, so this matches the file's own precedent.

The cap counts what the parent **ends up with**, not what this plan adds — caught by the commit-gate review. The earlier gate lets a parent through while it is still under the cap, so 2 existing + 3 planned would reach 5 under a cap of 3. A recovering run is the exception: its plan re-proposes the children that already exist and they are deduped by `idempotencyId`, so summing them there would refuse a resume that adds nothing.

## Verification

- `npx tsc --noEmit` clean; `npx vitest run src/automation/ src/core/` — **848 passed**.
- Each behaviour revert-checked: removing the `service.ts` line, the cap, or the projected-total rule each fails its own test.
- Four cases covered: over-cap refuses and falls through; exactly-on-cap still decomposes; existing children count toward the cap; a recovering run does not double-count.
- Commit gate ran twice — round 1 found the existing-children gap, round 2 `APPROVE`.

## Note on the new test file

`runnerExecution.coverage.test.ts` was already 7 lines under the 1500-line pre-commit cap, so the tests go in `runnerExecution.decomposition.test.ts` instead. Shaving prose to fit has now failed twice on other files in this session, and the coverage file returns to its exact `main` content. Its harness is copied rather than imported because `vi.mock()` is hoisted per file — the same reason `overlap.test.ts` and `integration.coverage.test.ts` each carry their own — and the copied-but-unused imports and fixtures were stripped.

Closes AGT-4122.

### DIFF
--- a/src/automation/decompositionLimits.test.ts
+++ b/src/automation/decompositionLimits.test.ts
@@ -1,0 +1,57 @@
+import { describe, it, expect } from 'vitest';
+import { plannedNewChildren, refuseForChildCap } from './decompositionLimits.js';
+
+const scope = (over: Partial<Parameters<typeof plannedNewChildren>[0]>) => ({
+  existingChildren: 0,
+  recovering: false,
+  plannedChildren: 1,
+  ...over,
+});
+
+describe('plannedNewChildren', () => {
+  it('charges a fresh plan for everything it proposes', () => {
+    expect(plannedNewChildren(scope({ plannedChildren: 4 }))).toBe(4);
+  });
+
+  it('charges a recovering plan only for the surplus it adds', () => {
+    // Re-proposed children are deduped by idempotencyId downstream, so summing
+    // them would spend budget on issues that already exist.
+    expect(plannedNewChildren(scope({ existingChildren: 3, plannedChildren: 4, recovering: true }))).toBe(1);
+  });
+
+  it('never goes negative when a recovery re-proposes fewer than exist', () => {
+    expect(plannedNewChildren(scope({ existingChildren: 5, plannedChildren: 2, recovering: true }))).toBe(0);
+  });
+});
+
+describe('refuseForChildCap', () => {
+  it('admits a plan that lands exactly on the cap', () => {
+    expect(refuseForChildCap(scope({ plannedChildren: 3 }), 3)).toBeNull();
+  });
+
+  it('refuses a plan that would leave more children than the cap', () => {
+    expect(refuseForChildCap(scope({ plannedChildren: 4 }), 3)).toMatch(/4 sub-issues.*over the 3 cap/);
+  });
+
+  it('counts children the parent already has toward the cap', () => {
+    // The pre-check upstream only asks "is the parent already at the cap?", so
+    // 2 existing + 2 planned slipped through and left 4. (AGT-4122)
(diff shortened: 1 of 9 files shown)


unohee:
## 🔍 Fresh review of e28db14 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The full decomposition config is now wired from service startup into the runner, and the child-cap projection correctly rejects plans that would exceed the parent cap. However, daily-limit enforcement still only checks whether one slot is available before planning, then creates and registers the entire plan, so it does not enforce the configured maximum.

unohee:
## 🔍 Fresh review of e28db14 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The decomposition object is now forwarded and child-cap projection is enforced, but the daily cap remains a per-decomposition admission check rather than a hard daily issue budget. A plan can be admitted with remaining capacity and then create enough children to exceed the configured daily limit.

unohee:
## 🔍 Fresh review of 56359c0 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The projected-cap checks work for a single decomposition and the production configuration is now forwarded to the runner. They are not atomic across concurrent pipeline runs, so the daily limit can still be exceeded under the supported parallel runner mode.

unohee:
## 🔍 Fresh review of ea0b52d (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The decomposition reservation accounting is now correctly separated from persisted creation counts, including the historical restart regression. However, the new service-instance lock is not released when shutdown cleanup fails, leaving the process permanently unable to restart.

unohee:
## 🔍 Fresh review of 8dd44a4 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The autonomous runner now enforces projected child and daily caps atomically, and the reservation persistence defect from the prior review is resolved. However, the production `/api/plan/dispatch` path still invokes the same child-creation primitive without either admission check or a reservation, so configured decomposition limits remain bypassable through the dashboard.

PR #474, opened 2026-08-30
fix(reconciler): report when a fenced run actually frees (AGT-4126)

unohee:
## What

`[Reconciler] Keeping <id> fenced until its original executor exits` named a condition that **cannot be observed** once a container restart replaced the process holding the claim. The line described an endless wait and pointed away from the mechanism that actually resolves it.

It does resolve: `reconcileAbandonMs` (default = `leaseMs` = 10 min) frees the row by age — the container pid-reuse fallback from AGT-4053. Measured on vela today: 7 rows orphaned by a redeploy, and AX-1074 / AX-1076 crossed the threshold and returned to `EXECUTING` on their own.

## Why it matters

The wording misled a reader **with full source access** into diagnosing a healthy, self-healing system as permanently wedged. Every supporting observation was true — `lease_expires_at` NULL on all fenced rows, the fence keyed on claim *presence* rather than liveness, the owner instance `7-e68cb63e…` holding no live row anywhere in the ledger — and the conclusion was still wrong, because the age sweeper lives in a different file.

An operator who reaches that conclusion redeploys, resetting a recovery that would have finished in under ten minutes.

## What changed

- `DurableRunCoordinator.reconcileAbandonDeadline()` reads the deadline off the **same `updatedAt`** the sweep compares, so the reported wait and the actual free cannot drift.
- `formatFenceWait()` / `fenceWaitMessage()` live next to the policy they describe.
- The message names **both** exits the sweep has — age, and a pid probe showing the owner gone — and deliberately offers no renewal alternative: reaching `NEEDS_RECONCILE` already required a full lease of silence, and nothing renews a row in that state. (Caught by commit-gate round 1, which flagged the first draft's "unless the owner renews" as a transition the state machine does not have.)

**Fencing behaviour is unchanged** — only the message.

## Verification

- 47 coordinator tests + 664 across `src/automation/`, `tsc --noEmit` clean
- Mutation-checked: drifting the deadline off `leaseMs*2` fails 2 tests; restoring the old wording fails 2 tests
- Commit gate: `✓ APPROVE` — *"the reported deadline is derived from `updatedAt`, matching the age-sweep predicate in `reconcile()`"*
- Message asserted through a pure formatter rather than the `AutonomousRunner` harness, which is flaky by global-singleton and timer contamination

Closes AGT-4126.

### DIFF
--- a/src/automation/autonomousRunner.ts
+++ b/src/automation/autonomousRunner.ts
@@ -1246,7 +1246,11 @@ export class AutonomousRunner {
       // one place it actually protects something. (AGT-4094)
       const task = taskById.get(run.issueId);
       if (run.ownerInstanceId || run.leaseToken) {
-        console.warn(`[Reconciler] Keeping ${run.identifier ?? run.issueId} fenced until its original executor exits`);
+        // Not "until its executor exits": a container restart replaces the
+        // process holding the claim, so that exit is never observed and the
+        // line reads as a permanent wedge. What actually frees the row is the
+        // age sweep in durableRunCoordinator, so report its deadline. (AGT-4126)
+        console.warn(this.durableRuns.fenceWaitMessage(run));
         continue;
       }
 
--- a/src/automation/durableRunCoordinator.test.ts
+++ b/src/automation/durableRunCoordinator.test.ts
@@ -4,7 +4,7 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import type { PipelineResult } from '../agents/pairPipeline.js';
 import type { TaskItem } from '../orchestration/decisionEngine.js';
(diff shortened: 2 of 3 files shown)


unohee:
## 🔍 Fresh review of 85a6a6f (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Reviewed the committed diff and production wiring in `src/automation/autonomousRunner.ts` and `src/automation/durableRunCoordinator.ts`. The reported deadline derives from the same `updatedAt + reconcileAbandonMs` condition used by the reconciliation sweep, and the reconciler invokes the new coordinator-owned message; focused tests cover formatting, identifier fallback, and deadline calculation.

PR #475, opened 2026-08-30
fix(ledger): count metrics within the daemon's dispatch scope (AGT-4127)

unohee:
## What

`getMetrics` counted the whole `automation_runs` table. That table is a permanent record: it still holds 158 rows (of 254) referencing macOS paths from before containerisation, plus projects disabled since. Those rows can never execute, but they dominated the totals — `RETRY_AT` read 88 where the live figure was 48, `READY` read 24 where it was 4. Both misled this session's monitoring into a wrong diagnosis before scoping by hand.

## How

- `DurableRunCoordinator.getMetrics(now, allowedProjects?)` scopes `byState` and `expiredActiveLeases` through **`isAllowedProjectPath` — the same predicate admission uses** — so the counts cannot describe a different project set than dispatch does.
- Out-of-scope rows are reported as one `outOfScope` number, not silently dropped.
- Omitted scope = raw table (explicit caller choice). Empty allow-list = the daemon's "no restriction" config, which decisionEngine reads as allow-everything — so everything is in scope. The two are deliberately distinct (gate round 1).
- `effectsByStatus` / `openCircuits` stay global: keyed by their own ids, not project paths; documented instead of half-scoped.
- `AutonomousRunner.getStats()` passes its live allow-list.
- No deletion — the ledger stays the recovery record (per AGT-4124, parked branches still hold real work).

## Review rounds

1. Gate: empty-vs-omitted conflation + sibling metric unscoped → both fixed, tested.
2. Gate: the expired-lease test asserted `<=` on counts that were both zero — replaced with two held EXECUTING runs probed past their lease, asserting exactly 2 raw / 1 scoped / 1 outOfScope.
3. Gate: ✓ APPROVE.

## Verification

- 52 coordinator tests, 860 across `src/automation/` + `src/orchestration/`, `tsc --noEmit` clean
- Mutation-checked: exact-match instead of the shared predicate → 1 test fails; scoping skipped → 1 test fails; sibling metric left global → 1 test fails

Closes AGT-4127 (scope-fencing half; Linear terminal reconciliation is follow-up work on the same issue).

### DIFF
--- a/src/automation/autonomousRunner.ts
+++ b/src/automation/autonomousRunner.ts
@@ -28,7 +28,7 @@ import {
   type TaskState,
   type ProjectInfo,
 } from './runnerState.js';
-import { taskEventKey, DecisionEngine, DecisionResult, TaskItem, getDecisionEngine, classifyStuck } from '../orchestration/decisionEngine.js';
+import { taskEventKey, DecisionEngine, DecisionResult, TaskItem, getDecisionEngine, classifyStuck, effectiveProjectScope } from '../orchestration/decisionEngine.js';
 import { getCoordinationStore } from '../coordination/coordinationStore.js';
 import { OPERATOR_PARK_REASON, shouldReadmitEarly } from '../coordination/operatorAnswers.js';
 // ExecutorResult used via execution.reportExecutionResult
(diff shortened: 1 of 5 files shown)


unohee:
## 🔍 Fresh review of bf7319c (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Scoped coordinator metrics correctly share DecisionEngine's path-containment predicate and tests cover empty, raw, nested-path, and expired-lease cases. However, production wiring scopes against configured allowed projects rather than the runner's enabled-project selection, so disabled repositories remain counted as live daemon work.

unohee:
## 🔍 Fresh review of 941148d (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The scoped ledger wiring is present and the empty/undefined scope distinction is correctly exercised. However, scoped ledger membership disagrees with dispatch on case-insensitive platforms, so admitted work can be reported as out of scope.

PR #495, opened 2026-08-30
fix(cli): report why Linear auth failed instead of claiming it is "not configured"

unohee:
## Summary

`openswarm work` on a machine with a revoked Linear OAuth token reported **"Linear is not configured"** and told the operator to set `linearApiKey` — a setting that was already present, and independently dead. Linear *was* configured; its refresh was rejected.

`ensureValidToken` already builds an accurate message. `ensureTaskSource` ended in a bare `catch { return null; }` that threw it away, collapsing four distinct causes into one `null` that the caller could only render one way.

## What changed

- `resolveTaskSource()` returns the cause alongside the source: `unconfigured` / `credential-rejected` / `transient`.
- `ensureTaskSource()` remains a thin wrapper, so the six existing call sites in `reviewCommand.ts` and `reviewMaxCommand.tsx` are untouched.
- `TokenRefreshError` carries the HTTP status structurally instead of callers parsing it out of a message.
- `openswarm work` renders guidance per cause, and suppresses its own re-auth line when the provider's message already carries one.

## Why an allow-list, not the 4xx range

Classification treats **400/401/403** as a dead credential and everything else as transient. The first draft used `status >= 400 && status < 500`; review caught that **429 is ordinary rate limiting** (and 408 a timeout), and reading either as a dead credential tells the operator to re-authenticate and discard a grant that still works.

The asymmetry is the point: a wrongly *transient* verdict costs one retry, a wrongly *terminal* one costs a working token.

## Verification

- Measured against the actually-revoked local token, before → after:
  ```
  - Linear is not configured — `openswarm work` picks issues from your tracker.
  - Run `openswarm auth login --provider linear` (or set linearApiKey + linearTeamId ...)
  + Linear rejected the stored credential — it is configured, but no longer valid.
  +   Token refresh failed (400): {"error":"invalid_request","error_description":"Refresh token revoked"}. Run: openswarm auth login --provider linear
  ```
- `src/auth` + `src/cli`: 45 files, 770 tests passed.
- Focused `reviewCommand` + `workCommand`: 97 tests passed (9 new, covering all three causes plus 429/408/404/422 and the non-Error throw).
- typecheck, lint, build: passed.
- `openswarm review`: REVISE (the 429 defect) → fixed → **APPROVE**.

Closes AGT-4148. Same class of defect as AGT-4012.

### DIFF
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -2,7 +2,7 @@
 // OpenSwarm - Auth Module
 // ============================================
 
-export { AuthProfileStore, ensureValidToken, type AuthProfile } from './oauthStore.js';
+export { AuthProfileStore, ensureValidToken, TokenRefreshError, type AuthProfile } from './oauthStore.js';
 export {
   runOAuthPkceFlow,
   loginAndSaveProfile,
--- a/src/auth/oauthStore.test.ts
+++ b/src/auth/oauthStore.test.ts
@@ -287,3 +287,47 @@ describe('ensureValidToken', () => {
     await expect(ensureValidToken(new AuthProfileStore(), 'gpt:default')).rejects.toThrow(/auth login/);
   });
 });
+
+// The CLI classifies a dead credential by the *shape* ensureValidToken throws:
+// name 'TokenRefreshError' plus a numeric HTTP status. Those classifier tests
+// build that shape by hand, so without this the two halves of the contract could
+// drift apart silently. (AGT-4148)
+describe('ensureValidToken refresh failures', () => {
+  const expiredProfile = () =>
+    validProfile({ provider: 'linear', expires: Date.now() - 60_000 });
+
+  async function refreshWith(status: number, body: string) {
+    writeStore({ 'linear:default': expiredProfile() });
+    const { AuthProfileStore, ensureValidToken, TokenRefreshError } = await loadModule();
+    vi.stubGlobal(
+      'fetch',
+      vi.fn(async () => new Response(body, { status })),
+    );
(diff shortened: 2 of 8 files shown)


unohee:
## 🔍 Fresh review of ad79abe (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The new failure classification is correctly surfaced by `openswarm work`, but the standalone `openswarm review --issues` path still discards it and emits the old ambiguous remediation. Unit coverage verifies the classifier and injected work-command results, not this default production path.

unohee:
## 🔍 Fresh review of 67ae154 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the committed diff and production wiring for both `review --issues` and `work`: failed Linear initialization now preserves and surfaces unconfigured, rejected-credential, and transient causes, while successful paths retain the prior task-source behavior. `TokenRefreshError` carries the HTTP status from the refresh endpoint and the classifier deliberately avoids mislabeling 408/429 or arbitrary 4xx responses as revoked credentials.
  Suggestions:
    - Add an oauthStore test that asserts a non-2xx refresh rejects with `TokenRefreshError` and preserves its `status`; the CLI classifier tests currently construct the error shape manually.

unohee:
## 🔍 Fresh review of 76389ac (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the full task-source resolution/error-classification wiring in `reviewCommand.ts` and `workCommand.ts`, plus the refresh-error production path and updated tests. Rejected OAuth refreshes preserve HTTP status and produce distinct re-auth guidance; unavailable and transient paths remain separately handled.

PR #497, opened 2026-08-31
fix: publish a draft PR when a run parks terminally for a human

unohee:
## Summary

A run that stops because it needs the operator is not a failure to hide — it is the run having built as far as it can and arrived at the point where a human has to look. That is exactly when the work should become reviewable.

It wasn't. All three terminal parks committed the partial work to a **local** branch and deleted the worktree without ever pushing:

- sandbox infeasibility (`DoD appears unsatisfiable`)
- the rejection limit (`Reviewer rejected 3 attempts`)
- retry exhaustion (`Autonomous execution failed 4 times`)

Measured on the deployed daemon's ledger: **14 `NEEDS_HUMAN` runs, every one holding a branch, not one PR between them.**

`publishOnPark.ts` (AGT-4076) already did this for runs parked on an `ask_human` question, but its trigger is `finalStatus === 'waiting_on_operator'`, which none of the three terminal parks produce.

## What changed

- **`publishStuckWork`** — opens a **draft** PR for a terminally parked run, with the park reason in the body. Draft because nothing reviewed the work and it is very likely incomplete; a ready PR would put it through CI.
- **Publication is a pre-removal hook of `removePreservedWorktreeAt`**, not a call before it. That way it inherits the worktree lifecycle lock and the live-owner re-check that already guard removal, so a push can't race a worker that resumed the tree, and it runs after the pre-cleanup WIP commit. Detached HEAD skips publication rather than pushing the wrong branch.
- **Unlike the operator-question path, this publishes uncommitted work too.** That path passes `committedOnly` because the run resumes from the tree; this tree is about to be deleted, and the WIP commit that would otherwise capture the remainder swallows its own failures.
- **A publication failure is logged and removal proceeds.** Cleanup cannot depend on GitHub being reachable, and the park stands either way.
- **The PR URL goes into the tracker comment** the operator actually reads.
- **Worker prompt (en + ko):** build and commit everything the missing input does not block *before* calling `ask_human`. Publication on that path is committed-only, so uncommitted work at the moment of the question was being discarded rather than published.
- `worktreeManager.ts` sat exactly on the 1500-line pre-commit cap, so the `gh` invocation and the duplicate-issue PR guard move to `ghPullRequests.ts` (pure relocation).

## Known consequence

Every stuck run now leaves a PR whose body contains `Closes <ID>`. If an operator reopens that issue and a later run completes it, `findDuplicateIssuePRs` will see the stuck draft and open the good PR as a draft too. That is the duplicate guard behaving as designed, and it is already true of the AGT-4076 operator-question path — closing the stuck draft when reopening the issue is the intended workflow. Flagging it because it is a real behavioural change, not a silent one.

This does not backfill the 14 branches already stranded; they are in other repositories and pushing them is an operator decision.

## Verification

- `npx tsc --noEmit` clean
- Full suite: **5206 passed**, 10 skipped, 0 failed
- New tests: 6 for `publishStuckWork` (draft options, park reason in body, no-commits and network-failure paths, declared vs inferred file scope), 4 for the cleanup hook (runs with the tree intact and the WIP commit already made; skipped for a resumed tree; removal still proceeds when publication throws; detached HEAD skipped)

## Review gate

Tier 1 **APPROVE** (Codex gpt-5.6-terra) on the substantive change, then a valid **REVISE** (the new module was untracked) which is fixed here. The re-review timed out at 300s, and the `claude` adapter fallback failed to start, so the final pass over the `committedOnly` change and the module extraction was **tier 3 inline self-audit**. Worth a careful look at `publishOnPark.ts` and the hook ordering in `worktreeManager.ts`.

---


unohee:
## 🔍 Fresh review of 56755fc (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Terminal paths are correctly wired through the lifecycle-locked pre-cleanup hook and draft publication, but the new remote-publication path can push binary data that the normal publication guard deliberately excludes. Add the same staging guard to the pre-cleanup WIP commit (or otherwise ensure those files cannot reach the pushed branch) and cover the terminal cleanup path.

unohee:
## 🔍 Fresh review of b49f072 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The terminal publication path is wired, but it bypasses the durable-run publication fence and artifact attachment used by both existing publication paths. That permits a stale executor to push after its durable claim has ended and leaves a successfully created draft PR absent from the run ledger.

unohee:
## 🔍 Fresh review of bb12712 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Terminal-park publication is wired through the lifecycle lock and has coverage for WIP commit ordering, owner races, cleanup failures, and binary-data refusal. However, the prompt collection bounding remediation remains incomplete in both locales: several untrusted arrays are joined in full before the per-string truncation runs.

unohee:
### Review gate disposition

Three `pr review --fresh` rounds. Two findings fixed, one out of scope.

**Round 1 — binary data on a pushed branch.** Valid, fixed in `b49f072`. The pre-cleanup WIP commit deliberately runs no binary guard, because unstaging a `.duckdb/.parquet/.pkl/.pt` there would strip it from the only copy that outlives the deleted worktree. Safe while the branch stayed local; not safe once it is pushed. Gated the *push* on the branch instead and fail closed when the base cannot be resolved — such a run keeps its work locally exactly as before and simply is not published.

**Round 2 — publication bypassed the durable fence.** Valid, fixed in `bb12712`, and the sharper form of it is right: the worktree lifecycle lock proves no worker is still editing the tree, which is **not** the same guarantee as this executor still owning the run. `markNeedsHuman` already is that proof — it refuses any row in a non-parkable state or still carrying an owner or lease — and all three terminal paths were discarding its return value. Publication is now gated on it. Cleanup still runs when it is false; only the push is withheld.

The ledger attachment half of that finding is **acknowledged and not fixed**, deliberately. No ledger method accepts a publication on a `NEEDS_HUMAN` row (`attachPublication` needs a live claim, `recoverPublishedRun` only accepts `NEEDS_RECONCILE`/`WAITING_EXTERNAL`), and writing `pr_url` onto a parked row feeds the reconcile classification that produced the phantom rows which idled the whole loop on 2026-08-29. That is a durable state-machine change deserving its own PR and its own reasoning about AGT-4145's head-identity contract. The reasoning is recorded in `publishOnPark.ts` so it is not silently forgotten. Until then the operator finds the PR through the tracker comment, and a later reviewed publication reuses it by branch name.

**Round 3 — out of scope.** The finding is that untrusted arrays are joined in full before per-string truncation in both locale prompt files. Those joins (`en.ts:114-207`, `repo.workspaces.join(', ')` and friends inside `promptDataBlock`) are unchanged on `origin/main`; this PR's entire prompt diff is one prose bullet per locale, 1 insertion and 1 deletion per file. Real observation, separate remediation — flagging it here so it is not lost rather than folding it in.

Round 3 produced no new finding about this change, which is the stop condition.

unohee:
## 🔍 Fresh review of bb12712 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Terminal-stuck publication is wired through the lifecycle lock and durable park fence, and the binary-data branch guard is exercised. However, the prompt changes still permit multi-megabyte untrusted contexts because per-item limits are not an aggregate prompt budget.

unohee:
Independent fresh review round returned `REVISE`, on one finding: *"the prompt changes still permit multi-megabyte untrusted contexts because per-item limits are not an aggregate prompt budget."*

**The finding is true, and it is not this PR's.** Verified rather than assumed:

- This PR's entire prompt diff is **one prose line per locale file** (`git diff origin/main...HEAD -- src/locale/prompts/{ko,en}.ts`): telling the worker to build and commit the unblocked part before calling `ask_human`. No code, no assembly, no limits.
- The 10 changed files contain no prompt-assembly or budgeting code.
- The mechanism the reviewer describes is on `origin/main` already: `MAX_PROMPT_DATA_CHARS = 20_000` is applied **per item** by a helper in both locale files, with no aggregate cap across items, so N items still compose without a ceiling.

Merging on that basis: CI is 12/12 green, and the two findings the earlier rounds raised against *this* change — the unsafe-binary push and the missing durable ownership fence — are both fixed in `b49f072` and `bb12712`.

The aggregate-budget gap is filed separately so it does not get lost. Tracked work for this PR: AGT-4150.

PR #500, opened 2026-08-31
fix(coordination): stop asserting that sibling asks share a blocker

unohee:
## Problem

When an operator answers one question, `humanQuestions.ts` fans the answer out to every other open ask on that task and labels each one:

> Answered via a differently-worded ask for the same blocker

**The code cannot know that.** The paging gate directly above it says so in its own comment:

> This does mean a genuinely new, unrelated question from the same task while one is still outstanding will not page either - accepted for now; distinguishing "reworded" from "unrelated" needs more than a text diff and is not attempted here.

So the label states as established the exact thing that was explicitly not established. When the asks really are unrelated, the agent is told its question was answered on its own terms - with an answer to something else.

## Change

Labels only. The mechanism is untouched: the fan-out still settles siblings, which is what lets `openQuestionCount` reach zero and unparks the run.

- en: `Settled by the operator answer to another open ask on this task - check it applies before acting on it`
- ko: `이 작업의 다른 열린 질문에 대한 운영자 답변으로 함께 처리됨 - 이 질문에 해당하는 답인지 확인하고 쓰라`

This keeps the harness to the standard it already sets for its agents. AGT-4012 and AGT-4148 fixed the same shape of defect: reporting a cause that was never established.

## What I checked before changing anything

Three earlier concerns in this area turned out to be already handled, recorded so the next reader does not re-derive them:

- Questions 2..N on a task are **not** stranded by the no-re-page rule - the answer fan-out settles them (`humanQuestions.ts:238-260`).
- The agent is **not** blind to the grouping - `ResolvedHumanAnswer.questions` carries every question a single answer settled, and `formatAuthoritativeOperatorFeedback` renders them all above one `Operator answer:` line.
- `openQuestionCount` already counts suppressed asks correctly.

Only the label was wrong.

## Verification

- `src/locale` + `src/coordination`: 313 pass, 1 fail - **pre-existing**, proven by stashing this patch and re-running against pristine `origin/main` (`9704ffe`), where it still fails. It is an ANSI-colour assertion: the child process emits a yellow-wrapped `true`, and the test asserts the plain substring. It only bites where stdout is a TTY, which is why CI is green. Filed separately.
- 3 new tests pin the label; **mutation-checked** - restoring the old string turns all 3 red.
- typecheck, lint (0/0), build pass. `openswarm review`: **APPROVE**.

### DIFF
--- a/src/locale/en.ts
+++ b/src/locale/en.ts
@@ -372,7 +372,7 @@ export const en: LocaleMessages = {
       operatorPaged: 'Operator paged on Discord',
       humanAnswered: 'Human answered the blocking question',
       supervisorAnswered: 'Project supervisor answered the blocking question',
-      siblingAnswered: 'Settled by the operator answer to another open ask on this task — check it applies before acting on it',
+      siblingAnswered: 'Settled by the answer to another open ask on this task — check it applies before acting on it',
     },
     periodicReview: {
       title: 'Periodic repository review',
--- a/src/locale/en.ts
+++ b/src/locale/en.ts
@@ -372,7 +372,7 @@ export const en: LocaleMessages = {
       operatorPaged: 'Operator paged on Discord',
       humanAnswered: 'Human answered the blocking question',
       supervisorAnswered: 'Project supervisor answered the blocking question',
-      siblingAnswered: 'Answered via a differently-worded ask for the same blocker',
+      siblingAnswered: 'Settled by the operator answer to another open ask on this task — check it applies before acting on it',
     },
     periodicReview: {
       title: 'Periodic repository review',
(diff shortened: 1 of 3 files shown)


unohee:
## 🔍 Fresh review of c39b958 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The revised label is production-wired through `answerHumanQuestion`, but it is inaccurate for supervisor-originated answers. The sibling fan-out runs regardless of `actorRole` (`src/coordination/humanQuestions.ts:221-260`), while both revised locale labels assert the answer came from an operator (`src/locale/en.ts:375`, `src/locale/ko.ts:375`).

unohee:
## 🔍 Fresh review of 0e41ad6 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the production caller in `src/coordination/humanQuestions.ts:238-259`: every other open question on the task is settled with this label regardless of semantic equivalence or answer actor, so the revised EN/KO wording accurately describes the behavior and preserves the applicability warning. The locale-contract tests cover both catalogs and the required/non-asserted wording; no material defect found.

PR #501, opened 2026-08-31
feat(cli): reach the coordination plane from the terminal

unohee:
## What

The dashboard's orchestration, chat and threads pages drive four coordination endpoints. `attach` (AGT-4058) covered the two writes, so an operator could answer a parked question but had no way to *see* that one was waiting — the board was readable only in a browser.

| Endpoint | CLI before | CLI after |
| --- | --- | --- |
| `POST /api/coordination/message` | `attach` (file mandatory) | `attach` (file optional) |
| `POST /api/coordination/attachment` | `attach` | `attach` |
| `GET /api/coordination` | — | `openswarm board` |
| `GET /api/coordination/threads` | — | `openswarm threads` |

- **`openswarm board`** leads with the questions waiting on an operator and prints each `correlationId`, because that is what `attach` needs to answer the right exchange.
- **`openswarm threads`** defaults the repository the route requires to the working directory.
- **`attach` no longer requires a file.** Answering a question usually means sending text and nothing else, which the command could not do.
- **`OPENSWARM_DAEMON_HOST`** points any of this at a daemon on another machine. Three call sites had grown their own `http://127.0.0.1:${port}`; they now share `daemonEndpoint.ts`, which validates the host instead of pasting it into a template literal. Loopback stays the default — a remote host sends operator answers and attachment bytes off the machine, so it is opt-in.

## Verification

Run against a live remote daemon, not just mocked:

- `board` and `threads` both render; `board --json | jq` pipes cleanly (config logging goes to stderr).
- A bad host, an unreachable daemon and a nonsense `--limit` each fail the way they claim to; exit codes 1 / 1 / 0.
- Gate: typecheck, oxlint, build clean. Full suite at baseline parity with `origin/main` — 9 pre-existing failures (7 TUI ANSI assertions, `coordinationTools`), zero new.

## Defects the review gate caught, not the tests

1. **`exec` would have auto-started the wrong daemon.** `startServiceAuto()` only ever starts a daemon on *this* machine. With a remote host configured, a failed health check would have spawned a local daemon nobody asked for and then polled the remote for 30s anyway. It now reports the unreachable remote instead.
2. **`--limit` was unvalidated before slicing.** It reaches `formatBoard` straight from `Number.parseInt`, and `slice(-limit)` mishandles both `NaN` (returns the whole board) and a negative count (slices from the wrong end).

A third came from following up on a reviewer aside: `localhost` and `[::1]` were classified as remote, which would refuse an auto-start that was perfectly valid.

`telemetry.ts`'s command allowlist is itself a guard test — adding a command without listing it there is a red test, and it caught both new commands.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -730,9 +730,13 @@ program
   .argument('<issueId>', 'Linear issue id or identifier (e.g. AGT-123)')
   .argument('[files...]', 'Local file path(s) to upload (optional when -m is given)')
   .option('-m, --message <text>', 'Message to send with the file(s)')
-  .action(async (issueId: string, files: string[], opts: { message?: string }) => {
+  .option('-c, --correlation <id>', 'Answer one exchange by the correlationId `openswarm board` prints')
+  .action(async (issueId: string, files: string[], opts: { message?: string; correlation?: string }) => {
     const { runAttachCommand } = await import('./cli/attachHandler.js');
-    process.exitCode = await runAttachCommand(issueId, files, opts);
+    process.exitCode = await runAttachCommand(issueId, files, {
+      message: opts.message,
+      correlationId: opts.correlation,
+    });
   });
 
 // openswarm board / openswarm threads — the read half of the coordination plane
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -726,13 +726,40 @@ program
 
 program
   .command('attach')
-  .description("Upload file(s) to a task's coordination inbox and notify its agent")
+  .description("Send a message and/or file(s) to a task's coordination inbox and notify its agent")
   .argument('<issueId>', 'Linear issue id or identifier (e.g. AGT-123)')
-  .argument('<files...>', 'Local file path(s) to upload')
+  .argument('[files...]', 'Local file path(s) to upload (optional when -m is given)')
   .option('-m, --message <text>', 'Message to send with the file(s)')
   .action(async (issueId: string, files: string[], opts: { message?: string }) => {
     const { runAttachCommand } = await import('./cli/attachHandler.js');
     process.exitCode = await runAttachCommand(issueId, files, opts);
+  });
+
+// openswarm board / openswarm threads — the read half of the coordination plane
+
+program
+  .command('board')
+  .description('Show the coordination board: questions waiting on you, then recent activity')
+  .option('-p, --port <port>', 'Daemon port', parseTcpPortOption, 3847)
+  .option('-r, --repository <path>', 'Limit to one repository')
+  .option('-n, --limit <count>', 'Recent events to show', (v: string) => Number.parseInt(v, 10), 20)
+  .option('--json', "Print the daemon's raw payload")
+  .action(async (opts: { port: number; repository?: string; limit: number; json?: boolean }) => {
+    const { runBoardCommand } = await import('./cli/coordinationHandler.js');
+    process.exitCode = await runBoardCommand(opts);
+  });
+
+program
+  .command('threads')
+  .description('List coordination threads for a repository (defaults to the working directory)')
+  .option('-p, --port <port>', 'Daemon port', parseTcpPortOption, 3847)
+  .option('-r, --repository <path>', 'Repository path (default: cwd)')
+  .option('-s, --status <status>', 'Filter by status: open | resolved')
+  .option('-n, --limit <count>', 'Maximum threads to list', (v: string) => Number.parseInt(v, 10))
+  .option('--json', "Print the daemon's raw payload")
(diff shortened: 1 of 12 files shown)


unohee:
## 🔍 Fresh review of f244905 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  Checked the new endpoint abstraction against the web authorization boundary and re-checked applicable historical prompt-handler findings. Remote daemon support is nonfunctional when the daemon is secured, and `exec` can still hang indefinitely in both auto-start and task submission paths.

unohee:
Passing on a review finding you probably have not seen: it surfaced when your in-progress changes were briefly in a shared checkout I ran `openswarm review` in, so the output went to my terminal rather than yours. Sorry for the noise that caused on your end — I have since moved my own work to an isolated worktree.

**`OPENSWARM_DAEMON_HOST` redirects mutating traffic, but no web token is ever sent.**

The daemon gates requests on `OPENSWARM_WEB_TOKEN`:

- `src/support/web.ts:100-101` — `hasValidWebToken()` reads `process.env.OPENSWARM_WEB_TOKEN`
- `src/support/web.ts:148,154` — it is the gate on the protected routes

None of the CLI callers on this branch send one. Grepping `origin/feat/cli-coordination-access` for `WEB_TOKEN|Authorization|Bearer|x-openswarm-token` returns **0 hits** in `daemonEndpoint.ts`, `attachHandler.ts`, `coordinationHandler.ts`, `promptHandler.ts`, and `providerCommand.ts`.

The awkward part is the interaction with line 1419:

```ts
const listenHost = process.env.OPENSWARM_WEB_TOKEN?.trim() || trustTailscale ? '0.0.0.0' : '127.0.0.1';
```

Setting the token is one of the two things that makes the daemon listen off-loopback at all. So the configuration this feature exists to reach — a daemon on another host — is largely the same configuration that requires a token, and `OPENSWARM_DAEMON_HOST=vela openswarm board` would come back 403 against it.

Worth deciding explicitly rather than leaving to discovery: read the token from the same env var and send it, or document that remote use requires the Tailscale-trust path instead. Either is fine; silently 403-ing is the one outcome that costs someone an afternoon.

Not blocking, not my PR — just relaying evidence rather than dropping it.

unohee:
## 🔍 Fresh review of d346b61 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The daemon endpoint refactor correctly adds token authentication to attach, prompt, board, and threads, but provider was left unauthenticated. A secured remote daemon therefore makes `openswarm provider` misreport its live state and reject provider switches despite `OPENSWARM_DAEMON_TOKEN` being configured.

unohee:
## 🔍 Fresh review of a6d122e (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The coordination commands, message-only attach flow, shared auth headers, and daemon-host validation are wired and covered at the handler level. However, the newly advertised remote-daemon path transmits both bearer-equivalent credentials and coordination content over plaintext HTTP, with no way to select HTTPS.

unohee:
## 🔍 Fresh review of f54cce3 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The new board advertises correlation IDs as the way to address a parked question, but `attach` accepts no correlation ID and silently selects the newest matching question; replies can therefore be attached to the wrong still-open exchange. Telemetry’s new read-merge-write sequence is still non-atomic across concurrent CLI/daemon first runs, so it does not guarantee a stable installation ID as claimed.

unohee:
## 🔍 Fresh review of 03d787a (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The remote-daemon wiring is functional and authenticated when configured, but it still permits plaintext transport of coordination messages and attachments without a token. Also, provider switching treats endpoint-configuration/security failures as daemon absence and silently persists a local override.

unohee:
## 🔍 Fresh review of b242f8b (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The remote-endpoint/auth wiring is consistently applied across board, threads, attach, provider, and exec, and the new command routes match the daemon APIs. Fix attach's handling of endpoint-configuration errors: it currently hides actionable safety/configuration failures as a local-daemon availability problem.

unohee:
## 🔍 Fresh review of de9ede3 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the committed CLI wiring, daemon endpoint/auth handling, coordination route contracts, explicit correlation routing, and affected tests. The current implementation correctly authenticates remote calls, prevents plaintext token transmission by default, preserves explicit correlation IDs through `attach`, and the board/threads routes are production-wired.

PR #502, opened 2026-08-31
fix(automation): stop re-admitting a parked run from its own footprint

unohee:
## Problem

A run parked on `NEEDS_HUMAN` was pulled back to `READY` whenever its Linear card sat in `Todo`, `In Progress` or `In Review`. The pipeline writes `In Progress` itself when it claims a task, and parking does not move the card back — so that test was true on the very next heartbeat for every park that leaves the card alone. Each cycle re-claimed, re-executed, re-published and re-parked, burning an attempt and holding a slot the whole time.

```
20  parked            RETRY_AT     NEEDS_HUMAN  2026-08-30 10:04:15
20  transition        PUBLISHING   RETRY_AT     2026-08-30 10:04:15
20  transition        EXECUTING    PUBLISHING   2026-08-30 10:04:14
20  claimed           RETRY_AT     CLAIMED      2026-08-30 10:00:05
18  operator_resumed  NEEDS_HUMAN  READY        2026-08-30 09:25:00   ← nobody acted
18  parked            RETRY_AT     NEEDS_HUMAN  2026-08-30 06:08:26
```

Counters reached this way: AX-1030 = 20, AX-1027 = 16, AX-873 = 15, AGT-4119 = 9. Four `operator_resumed` events fired on the current build, the most recent an hour before this change.

## Why `Todo` survives and `In Progress` does not

Across the live ledger, every parked run in a state that triggered the resume was in `In Progress`, and none was in `Todo`:

| card state | parked runs | max attempt |
| --- | --- | --- |
| Backlog | 15 | 21 |
| *(none)* | 12 | 19 |
| **In Progress** | **6** | **20** |
| Todo | **0** | — |

So the loop is entirely the pipeline reading its own writes back as an operator signal. `Todo` is the state only an operator puts a parked card in — `durableRunCoordinator` already documents it as "the autonomous operator-reopen surface", and the same file already warns that `In Progress` "may be owned by a human or another daemon".

## Change

Re-admission of a non-question park now requires an operator act: the card in `Todo`, or an explicit dispatch from the issue board / `work` CLI. Both answer-fenced `ask_human` paths are untouched — they read the durable trace, not the ring buffer.

The stuck-issue log line advertised `Todo / In Progress`; it now names `Todo` alone. `In Progress` still recovers a purely label-stuck issue via `classifyStuck`, but an issue with a durable `NEEDS_HUMAN` run is filtered out before reaching that check, so `Todo` is the one state that works on both paths.

## Tests

Replaces the test that asserted the old resume with two guards:

- a park stays parked while its card is `In Progress`, **and an unrelated task takes the freed slot**
- the same park still resumes on explicit dispatch

Mutation-checked: restoring `['Todo','In Progress','In Review']` turns the first guard red. The pre-existing *"resumes durable NEEDS_HUMAN when an operator moves a stuck issue to Todo"* test passes unchanged — the first draft of this fix broke it, which is what led to narrowing to `Todo` rather than removing the card check.

`typecheck` / `lint` clean; the three park suites pass 28/28. Full suite: 8 failing files, all reproduced on unmodified `origin/main` on this machine (7 TUI + `coordinationTools`) and green in CI.

Closes AGT-4155.

### DIFF
--- a/src/automation/autonomousRunner.operatorQuestionPark.test.ts
+++ b/src/automation/autonomousRunner.operatorQuestionPark.test.ts
@@ -323,17 +323,39 @@ describe('stop re-dispatching a repeatedly-unanswered ask_human (AGT-4042)', () 
     internal.durableRuns.close();
   });
 
-  it('still resumes an unrelated NEEDS_HUMAN park once Linear reopens it', async () => {
-    // The other half of the mutual-exclusivity fix: making the ask_human path
-    // ignore linearState must not cost the OTHER park types their own
-    // resume condition.
+  it('leaves an unrelated NEEDS_HUMAN park alone while its card sits In Progress (AGT-4155)', async () => {
+    // 'In Progress' is where THIS run put the card when it claimed the task,
+    // and parking does not move it back. Treating that level as "the operator
+    // reopened it" re-admitted the park on the very next heartbeat: it
+    // re-claimed, re-executed, re-published and re-parked, once per cycle,
+    // holding a slot the whole time. Observed in production at attempt 20.
     const internal = await makeRunner();
     internal.durableRuns.markNeedsHuman('AGT-1', 'Reviewer rejected 4 attempts: still failing lint');
-    const reopened: TaskItem = { ...TASK, linearState: 'In Progress' };
+    const stillParked: TaskItem = { ...TASK, linearState: 'In Progress' };
+    // The slot the park was holding has to go somewhere: an unrelated task in
+    // the same cycle must still be admitted.
+    const sibling: TaskItem = {
+      ...TASK, id: 'AGT-2', issueId: 'AGT-2', issueIdentifier: 'AGT-2', title: 'unrelated work',
+    };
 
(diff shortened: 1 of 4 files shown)


unohee:
## 🔍 Fresh review of 9e6cb07 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  The In Progress spin is prevented for durable NEEDS_HUMAN records, and the added tests cover that filter plus CLI-style explicit dispatch. However, this changes the documented Linear-board recovery contract without updating its production entry point: moving a durable stuck card to In Progress no longer resumes it.

unohee:
## 🔍 Fresh review of 2172083 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the current NEEDS_HUMAN admission path and its durable-ledger interaction. The change correctly limits non-question parks to Todo or explicit dispatch, preserves question-specific answer gating, and has regression coverage for In Progress parking, sibling admission, and explicit dispatch.

PR #505, opened 2026-09-01
fix(autonomy): CodeQL path fallback + agent-owned DoD coordination (v0.21.4)

unohee:
## Summary
- **CodeQL gate**: `findSystemExecutable` now checks `OPENSWARM_CODEQL_PATH` and `/opt/codeql` before PATH — survives `bash -lc` entrypoints that drop compose PATH (vela infra_error root cause).
- **ask_human scope**: tool description + worker prompts (en/ko) reserve `ask_human` for true operator decisions (secrets, production access, irreversible actions, policy). DoD/scope/approach disputes go through `coordination_publish` to the orchestrator peer: implement the largest sandbox-verifiable subset, propose the DoD revision with evidence.
- **Draft DoD**: completion-criteria guidance now requires sandbox-achievable-only criteria — no "needs human" in DoD; operator-only remainder noted in `suggestedApproach`.

## Ops companion (deploy-side, not in this PR)
- vela entrypoint switched to `bash -c` + explicit PATH export; sandbox-executor recreated on the 0.21.3 image; 53 stale RETRY_AT runs returned to READY.

## Test plan
- [x] securityAudit tests incl. new OPENSWARM_CODEQL_PATH/empty-PATH case (34)
- [x] prompts tests both locales (71)
- [x] full suite 5310 passed
- [ ] CI green


unohee:
Follow-up commit: gate no longer throws on 'partial' — Go extractor permanently rejects build-mode=none, so AGT-3841-style Go pipelines failed closed forever. Partial now passes with the skip visible on skippedCodeqlLanguages/detail (baseline and current skip the same languages, so the introduced-findings comparison stays sound).

unohee:
Follow-up `78f6028`:

- Tests: CodeQL `partial` is no longer `infra_error` (the previous assertion was the old fail-closed contract).
- `.venv.bak` is ephemeral so worker-scope does not fail on the runtime symlink.
- New CodeQL findings are logged with the actual rules.
- Orchestrator: `host_read_file` / `host_search_files` on the development-host checkout; `web_search` on; bash still withheld.
- `web_search` prefers SearXNG via `OPENSWARM_SEARXNG_URL` (vega-search).

Remote is `Intrect-io/OpenSwarm`.

PR #568, opened 2026-09-09
fix(verify): restore isolated Python verification on VELA (AGT-4223)

unohee:
CGF worktrees repeatedly stopped before pytest because local `.env`/`.dev.vars`, customer-data links, and nested virtualenv interpreters were rejected as escaping symlinks. Eight tasks parked after repeated attempts. Copying the virtualenv alone also left uv editable installs pointing at the original checkout.

Use the same exclusion policy before source validation and copy; omit private configuration and ignored external local-asset links without dereferencing them. Keep tracked and ordinary source-link containment. Disable checkout hooks in disposable verification worktrees. Independently copy command-local Python dependencies for HEAD/base and relocate plain-path editable `.pth` entries to the snapshot, retaining dependency CA bundles.

Validation:
- 12 environment regressions; the initial 8 tests all failed before the fix.
- Full suite: 5,545 passed / 10 skipped. Typecheck and build pass; lint has no errors (one unchanged warning in webTools.test.ts).
- OpenSwarm working-tree review: APPROVE; fresh PR review: APPROVE.
- VELA attested companion, CGF commit `a882db13ea2731ab0a5a35a2f80c282ba2d73b23`: Python/pytest import succeeds, project module path asserted inside disposable checkout, real full pytest **1,932 passed / 57 skipped**.
- Live baseline comparison: intentional HEAD-only failure yields `headStatus=fail`, `baseStatus=pass`, `newFailure=true`; baseline pytest also **1,932 passed / 57 skipped**.

Deployment and affected-task recovery will be verified after CI and the fresh PR review.

Fixes https://linear.app/intrect/issue/AGT-4223

### DIFF
--- a/src/sandboxExecutor/bubblewrap.ts
+++ b/src/sandboxExecutor/bubblewrap.ts
@@ -14,6 +14,7 @@ import {
   untrackCliProcessTree,
 } from '../adapters/processTree.js';
 import { linkedMainCheckoutOf } from '../security/gitWorktreeIdentity.js';
+import { isPrivateWorkspaceFile } from '../support/environmentFiles.js';
 import type { SandboxExecutionResult } from './protocol.js';
 
 export interface WorkspaceIdentity {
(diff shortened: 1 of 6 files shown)


unohee:
## 🔍 Fresh review of 3b9e82f (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the production wiring from verify snapshot creation/base comparison through isolated shared-path copying, Python editable-install rebasing, and bubblewrap secret masking. The shared policy is applied consistently, private files are filtered before copy/dereference, and the new tests cover tracked, ignored, shared-path, symlink, worktree, and base-comparison cases.

PR #569, opened 2026-09-09
fix: ignore pytest-local artifacts at publication fence

unohee:
## Problem
Verification runs that commit `pytest-local/<case>` scratch output were rejected by the publication-scope fence as out-of-scope source files. This parked the run with `publication_scope_mismatch` even though the files were ephemeral test artifacts.

## Change
Treat the entire `pytest-local/` tree as ephemeral in the shared artifact classifier used by WIP staging, resume filtering, purge, and the publication fence.

## Validation
- `npm test -- --run src/support/worktreeEphemeral.test.ts` (6 passed)
- `npm run typecheck`
- `openswarm review --path .` — APPROVE


unohee:
## 🔍 Fresh review of 6be5524 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Checked the shared classifier and its production callers in WIP staging/purge and publication-scope filtering. The anchored `pytest-local(?:/|$)` rule correctly covers the previously omitted descendants without matching similarly named paths, and the focused regression test exercises a child path.

PR #580, opened 2026-09-09
fix(async-ui): contain rejected commands and overlapping monitor requests — prevent unhandled failures and polling fan-out

unohee:
## Background

Async command and monitor flows do not consistently contain rejected work or enforce one active request. A timeout may race with an underlying fetch that continues while future intervals start additional requests.

## Included follow-ups

* Handle rejected planner executions in the /plan command flow (src/support/planCommand.ts:51)
* Prevent timed-out monitor fetches from allowing overlapping underlying requests (src/tui/hooks/useMonitor.ts:37-67)
* Validate or clamp the monitor polling interval before creating setInterval (src/tui/hooks/useMonitor.ts:67)

## Completion criteria

* Planner rejection is caught and converted into a controlled command result.
* Monitor polling aborts, drains, or otherwise accounts for timed-out fetches before another request starts.
* Poll intervals are validated and clamped to safe finite bounds before scheduling.

- **PR #579 (swarm/AGT-3492-fix-network-access-restrict-notification)** — 1 file(s): `package-lock.json`
- **⚠️ MERGED PR #578 (chore/release-0.23.0) — not in this branch's history, verify its fix wasn't dropped** — 1 file(s): `package-lock.json`

## Linear
Closes AGT-3417

---

### DIFF
--- a/src/support/planCommand.test.ts
+++ b/src/support/planCommand.test.ts
@@ -46,21 +46,73 @@ afterEach(() => {
 });
 
 describe('runPlanCommand', () => {
-  it('dispatches the approved plan to the daemon', async () => {
+  it('dispatches the approved sub-tasks on yes', async () => {
     mockedRunPlanner.mockResolvedValue(plannerResult([
-      { title: 'a', description: 'a', estimatedMinutes: 10, priority: 1 },
-      { title: 'b', description: 'b', estimatedMinutes: 20, priority: 2, dependencies: ['a'] },
+      { title: 'A', description: 'da', estimatedMinutes: 10, priority: 2 },
+      { title: 'B', description: 'db', estimatedMinutes: 15, priority: 3, dependencies: ['A'] },
     ]) as never);
-    const fetchMock = vi.fn(async () => ({ ok: true, status: 200, json: async () => ({ mode: 'pipeline', taskIds: ['1', '2'] }) }));
+    const fetchMock = vi.fn(async () =>
+      new Response(JSON.stringify({ mode: 'linear', parentIssue: { identifier: 'INT-9' } }), { status: 200 }),
+    );
     vi.stubGlobal('fetch', fetchMock);
 
     const { io, out } = makeIO(['yes']);
+    await runPlanCommand('build X', io, { projectPath: '/tmp/proj' });
+
+    expect(fetchMock).toHaveBeenCalledOnce();
+    expect(String(fetchMock.mock.calls[0][0])).toContain('/api/plan/dispatch');
+    const body = bodyOf(fetchMock);
+    expect(body.goal).toBe('build X');
+    expect(body.projectPath).toBe('/tmp/proj');
+    expect(body.subTasks).toHaveLength(2);
+    expect(out.join('\n')).toContain('INT-9');
+  });
+
+  it('does not dispatch on no', async () => {
+    mockedRunPlanner.mockResolvedValue(
+      plannerResult([{ title: 'A', description: 'd', estimatedMinutes: 5, priority: 2 }]) as never,
+    );
+    const fetchMock = vi.fn();
+    vi.stubGlobal('fetch', fetchMock);
+
+    const { io, out } = makeIO(['no']);
     await runPlanCommand('g', io, {});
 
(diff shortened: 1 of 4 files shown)


unohee:
## Independent review — DO-NOT-MERGE as it stands

CI is 8/8 green and that is real, but it is not informative for this diff. Three findings, each verified by running the code rather than reading it.

### 1. The "overlapping monitor requests" half is a no-op — it was fixed two months ago

`main`'s `src/tui/hooks/useMonitor.ts` already contains the entire containment this PR's title claims to add: the `inFlight` re-entry guard, `AbortController` + `activeController`, the 15s timeout calling `controller.abort()`, `Promise.race([fetcher(port, signal), onAbort])`, `finally { clearTimeout; inFlight = false }`, and `activeController?.abort()` on cleanup. Landed in `cf14da2`/`5877f3b` on 2026-07-02 (INT-2316). It reaches the wire too — `monitorApi.ts:12-14` composes the caller's signal via `AbortSignal.any`, and all four fetchers forward it.

Proof the new tests pin nothing: **revert `useMonitor.ts` to `main` and keep the new test file → 5 of 7 still pass, including both headline tests** (`aborts a timed-out fetch…` and `does not start a second request while one is in flight`). The 2 failures are `clampPollIntervalMs is not a function` — a missing export, not a behavioural difference. Keeping the helper but reverting only the wiring → **7/7 pass**. The one line of production behaviour this PR adds to that file is pinned by zero tests, and is unreachable: the sole caller `MonitorPanel.tsx:18` passes no interval, so the `= 5000` default always applies.

The test file is a useful characterization net for behaviour that already exists. It must not be presented as pinning a fix.

### 2. It deletes four working tests, and three production paths can now be broken silently

`planCommand.test.ts` is +26/-59. It removes `does not dispatch on no`, `drops a sub-task on edit, then dispatches the remainder`, `uses the single-task path when no decomposition is needed`, and weakens `dispatches the approved sub-tasks on yes`.

They were not deleted because they broke: **`main`'s original 5 tests pass 5/5 against this PR's own `planCommand.ts`.**

Mutations applied one at a time to the PR's production file:

| Mutation | PR suite | `main` suite |
|---|---|---|
| delete the `decision === 'no'` early return (`planCommand.ts:90-93`) | green | red |
| neuter the drop filter (`planCommand.ts:105`) | green | red |
| `if (false)` on the `mode === 'linear'` branch (`planCommand.ts:154`) | green | red |

Concrete consequence: a later change inverts the drop-index filter, `/plan` → `edit` → "drop 2" dispatches sub-task 2 anyway, the user's rejected work is queued to the daemon, CI stays green. The `edit` branch — index parsing, Set construction, filter, empty-after-drop cancel, re-render loop — is the most intricate logic in the file and is now completely untested.

### 3. The one real fix is good, but its stated motivation is false

The `try/catch` around `runPlanner` (`planCommand.ts:52-65`) is correct and its two tests genuinely pin it (both go red against reverted production). But the body's premise — "so the command surface never surfaces an unhandled promise rejection" — is not true: **both** call sites already catch (`ChatPanel.tsx:167-176` and `:234-245`). This is defensive hygiene, not a bug fix.

### Also worth knowing

`tsconfig.check.json` excludes `src/**/*.test.ts(x)`, so the green Type Check job never saw either new test file. The loose stubs in them — a `{ ok, status, json }` literal standing in for `Response`, and a `vi.fn` returning `Promise<void>` where `Promise<Table>` is declared — are unverified.

The `package-lock.json` churn in `gh pr diff` is a false alarm: computed against merge-base `d38b7d8`, but `git diff origin/main FETCH_HEAD -- package-lock.json` is empty. Zero lockfile change lands.

### Path to merge

1. Restore the four deleted `planCommand.test.ts` tests — they pass unmodified against this PR's production code, so this is a pure revert.
2. Drop the "overlapping monitor requests / polling fan-out" claim from the title and body; describe `useMonitor.test.tsx` as a regression net for existing behaviour, and record that AGT-3417's second bullet was closed by INT-2316 on 2026-07-02.
3. Either wire an interval through `MonitorPanel` or make `clampPollIntervalMs` module-private and drop its two unit tests.

With 1–3 this is a small, safe, net-positive PR. As it stands, merging it reduces `planCommand.ts` coverage in exchange for one genuine 14-line `try/catch`.

PR #583, opened 2026-09-09
[Bug] /api/health spikes past the 10s HEALTHCHECK timeout at concurrency 9, flapping the container unhealthy

unohee:
## TL;DR

At `maxConcurrentTasks: 9` with 6 tasks running, `/api/health` occasionally takes
longer than the container's 10-second HEALTHCHECK timeout, so the container flaps to
`unhealthy` while the daemon is in fact serving. Not a crash and not a stall of the
size first suspected — measured below — but the operator-facing endpoint does have
multi-second spikes under this concurrency.

## Measured on vela, 2026-08-29 18:45–18:53

From the **host**, direct HTTP (no docker exec), 8 consecutive samples:

```
116  245  40  99  122  173  191  198   ms
```

From **inside the container**, with a latency breakdown:

```
200  namelookup 0.00002s  connect 0.00018s  total 1.004 s
200  namelookup 0.00002s  connect 0.00016s  total 0.921 s
200  namelookup 0.00015s  connect 0.00033s  total 0.009 s
200  namelookup 0.00003s  connect 0.00021s  total 0.187 s
200  namelookup 0.00002s  connect 0.00133s  total 0.094 s
```

`connect` is instant every time, so the listener accepts immediately and the variance
is in the handler / event loop. Docker's health log for the same period:

```
18:45:12 exit=0   18:45:50 exit=0   18:46:23 exit=-1
18:47:03 exit=0   18:47:37 exit=-1        (exit=-1 = "exceeded timeout (10s)")
```

HEALTHCHECK is `curl -fsS http://localhost:3847/api/health`, timeout 10 s, retries 3.

## Two wrong turns worth recording, so they are not repeated

* **"The event loop is blocked for 29 s" — false.** That came from
  `docker exec … curl --max-time 30` returning 30,231 ms, which is the `--max-time`
  cap, not a measurement. Host-side probing never exceeds 245 ms.
* **The control used to "prove" it was invalid.** A timer-lag probe run in a second
  process reported 36 ms max over 574 samples and was read as "the OS is fine, so the
  daemon's own loop is blocked". That probe was itself started through `docker exec`
  and only measured lag *after* it started, so it could not have observed the window
  in question. `docker exec openswarm true` is 50 ms, so exec was never the cost either.

## Load context

Host load average 7.50 with several other containers; the daemon's cgroup peak is
6.0 GB of a 16 GB limit and CPU has been observed at 261 % of a 4.0-CPU quota. Memory
is not the constraint.

## Approach

* Give the health endpoint a path that cannot queue behind heartbeat work (serve a
  cached liveness snapshot, refreshed on a timer), so the probe measures "the process
  is alive" rather than "the loop is free right now".
* Failing that, raise the HEALTHCHECK timeout to exceed the observed spike and keep
  `retries`, so a single slow sample does not flip the container.
* Do **not** paper over a real regression: if spikes grow past a few seconds, find the
  synchronous work rather than widening the timeout again.

## DoD

- [ ] `/api/health` responds within the HEALTHCHECK timeout while the daemon runs its
      configured concurrency, measured over at least 30 consecutive probes
- [ ] The container does not report `unhealthy` during a normal heartbeat cycle
- [ ] Whatever produces the multi-second spikes is identified in the issue, not just
      timed out around

## Correction: this is not resource pressure, and the load framing was wrong

The operator pushed back — *"우리 서버 자원이 남아도는데 서버문제는 아닐걸 htop 찍어보면 알텐데"* —
and they are right. The issue's "Load context" section drew the wrong conclusion from a
load average I never divided by the core count.

Actual headroom on vela:

```
cores                       16
load1 / load5               11.17 / 9.00      -> ~70 % of capacity, not saturated
memory                      60 GiB total, 28 GiB available
top CPU consumer            qemu-system  381.8 %      (a VM, not OpenSwarm)
openswarm node              118.2 % CPU, 1.8 GiB RES
openswarm container         118 % of its 400 % quota, 1.87 / 16 GiB (11.7 %)
```

And per-thread, which settles it:

```
top -b -H -p <node pid>
  main thread (TID == PID)   60.0 %   R
  libuv workers x4            0.0 %   (2:42 cumulative each)
  51 threads total
```

The main thread is **not** pegged. So:

- The server is not the constraint — it has most of its capacity free.
- The container is nowhere near its own CPU or memory limit.
- The daemon's event loop is not continuously saturated either.

## What this changes

- **Retracted:** the recommendation to lower `maxConcurrentTasks` back to 6–7. It was
  based on "the box is loaded", which is false.
- **Retracted:** attributing the latency spikes to concurrency at all. Six concurrent
  tasks against a 60 % main thread does not explain a >10 s response.
- **Still true and still unexplained:** `connect` is instant while `total` varies
  0.009–1.0 s with rare excursions past the 10 s HEALTHCHECK timeout. Something on the
  main thread occasionally runs long. Candidates measured and **ruled out** so far:
  lifecycle-lock SQLite opens (51 opens = 13 ms total, 0.2 ms each) and the
  coordination board parse (1.1 MB, 7.5 ms).

The DoD item "whatever produces the multi-second spikes is identified in the issue, not
just timed out around" therefore stands, with two candidates now eliminated rather than
suspected.

### 2026-08-29T11:22:10.127Z
**Symptoms confirmed, two more mechanisms ruled out.**

Measured on vela (container `restarts=0`, so this is not a crash loop):

| probe | result |
|---|---|
| `/api/health` call 1 | **8.50 s** |
| `/api/health` calls 2, 3 | 0.10 s, 0.11 s |
| `/api/stats` | **19.98 s** (HTTP 200, 935 bytes) |
| Docker HEALTHCHECK (`curl /api/health`, 10 s timeout) | timed out 4/4 consecutive |

So the loop stalls in **bursts of 8-20 s**, not continuously.

**Ruled out this pass:**

1. *The `/api/stats` handler itself.* Its only `await` is `getAdapterSummary()`, whose `resolveAdapterDefaultModel` calls are short-circuited by `??` because the live config has `worker.model`/`reviewer.model` set. The 20 s is spent **before the handler runs** — the request sat in the queue.
2. *Knowledge-graph JSON cost.* `knowledge-graph/` is 570 MB across 707 files (largest 30 MB), and `loadGraph`/`saveGraph` do a synchronous `JSON.parse` / `JSON.stringify(..., null, 2)`. Measured on the real 20 MB `dev-openswarm.json` (10,908 nodes / 58,883 edges) inside the container: **read 62 ms, parse 74 ms, stringify pretty 93 ms, compact 61 ms.** Not the blocker. Hypothesis withdrawn.

**Still open, and the strongest remaining lead:** one knowledge scan in the last 3 h took **160,390 ms** — for a *smaller* graph (318 nodes / 317 edges) than the ones that finish in ~400 ms (481 nodes / 3,034 edges). Duration is not a function of graph size, so the time is in the walk or the git enrichment, not the serialisation. `SCAN_TIMEOUT_MS` is 30 s and bounds both the walk and the import parse, and `enrichWithGitInfo` issues one `git log`, not one per node — so none of the three known bounds explains 160 s. That gap is the next thing to measure.

Candidates ruled out across all passes so far: lifecycle-lock SQLite opens (13 ms/51), board parse (7.5 ms), the stats handler, JSON serialisation.

## Reproduced under a 200-sample probe, and narrowed to a 43-second silent window

Polled `/api/health` once a second for 200 samples (that handler is pure synchronous property reads ahead of both auth gates, so its latency *is* event-loop lag):

| bucket | samples |
|---|---|
| < 0.5 s | 196 |
| 0.5–2 s | 1 |
| 2–10 s | 1 |
| **≥ 10 s** | **2** |

avg 0.327 s. The two large ones: **28.58 s** at 12:05:04 UTC, and one at 12:06:41 that returned `30.00` — which is the probe's own `--max-time`, so read it as **≥ 30 s**, not as a measurement.

**Stalls are episodic and concurrency-linked.** The first 65 samples, taken while only 2 tasks were running, contained *zero* stalls. Both stalls landed minutes after three parked runs were released and the scheduler began dispatching them.

**The decisive new evidence:** around the second stall the daemon logged nothing between `12:06:32` and `12:07:15` — a **43-second gap with no output at all**, bracketing the ≥30 s measurement. Whatever blocks is inside that gap, which is why it leaves no log line.

## Ruled out by measurement (not by reasoning)

| candidate | evidence |
|---|---|
| `/api/stats` handler | its only `await` short-circuits; the 20 s was queue wait before the handler |
| knowledge-graph JSON | real 20 MB graph, in-container: read 62 ms, parse 74 ms, stringify 93 ms |
| knowledge scan | scans at 11:56 and 12:26 — 9 and 19 min from the stalls |
| embedding inference | measured: 203–378 ms per call, max loop lag 354 ms; model load 1403 ms / 779 ms lag. Real but an order of magnitude short, and no `[Memory]` line in the stall window |
| `findOpenPRFileOverlaps` preflight | cgf-portal has 4 open PRs; the `--json …,files --limit 1000` payload is **1,869 bytes**, fetch 478 ms, parse **0 ms** |
| lifecycle-lock SQLite opens | 13 ms across 51 opens (earlier pass) |
| coordination board parse | 7.5 ms (earlier pass) |

Seven candidates eliminated.

## Why this needs in-process instrumentation now

An external probe has run out of resolution. Every surviving candidate lives inside a window the log does not describe, and correlating from outside cannot narrow it further. The next step is `perf_hooks.monitorEventLoopDelay` (or a `SIGUSR1` inspector CPU profile) *inside* the daemon, sampled continuously and dumped when a threshold is crossed — that names the blocking frame directly instead of bracketing it.

Note for whoever adds it: the stall only appears under dispatch load, so the instrument has to be always-on and self-triggering. Reproducing it on a 2-task-deep loop will show nothing, as the first 65 samples of this probe did.

## Instrument shipped — PR #463 merged 2026-08-29 12:53Z

All 12 required checks SUCCESS. Both gates approved (`openswarm review` 3 rounds -> APPROVE; `pr review --fresh` -> APPROVE).

`src/support/eventLoopMonitor.ts`, armed in `AutonomousRunner.start()` before `engine.init()`. When a window's worst delay crosses 2 s it emits one line — which puts a marker *inside* the silent windows this issue is about — and classifies the stall from CPU accounting:

```
[EventLoop] Blocked 999ms  (mean 250ms over 1200ms) — cpu-bound: on-CPU 83.8% in user code (user 991ms) — synchronous JS [fs r0/w0 ctx v0/i979]
[EventLoop] Blocked 1014ms (mean 254ms over 1216ms) — blocked-off-cpu: off-CPU 99.8% (user 2ms, sys 1ms) — parked waiting, not computing [fs r0/w0 ctx v0/i16]
```

Both lines are from real stalls, not fixtures.

### Two corrections the build forced

**`perf_hooks.monitorEventLoopDelay` does not detect this.** It was the first implementation and a smoke test against a real block killed it: on Node v26 an 800 ms busy-wait left its `max` at the 21 ms warm-up value, and a 900 ms `Atomics.wait` left it at **0**. Replaced with timer lateness — the same technique that had already measured the 354 ms block one embedding call causes. Worth recording for anyone who reaches for that API here again.

**The classifier is narrower than first written.** It originally also branched on `fsRead`/`fsWrite` and voluntary-vs-involuntary context switches. Measured against real stalls those do not support it: macOS `getrusage` reports `fsRead: 0` across 12 `readFileSync` calls, and *involuntary* switches track being pre-empted **while busy** — 830 during a 1 s CPU burn versus 20 during a 1 s blocking wait, the opposite of the "starved by other tenants" reading they had. They now ride in the log line as raw evidence and decide nothing.

### What this does and does not settle

It does not name the blocking frame — by the time a delay is observable that frame has returned. It settles which **class** the stall belongs to, and those two classes have disjoint fixes: synchronous JS/CPU wants the work moved off the loop; parked-waiting wants the blocking call made async. Guessing wrong costs a full investigate-and-deploy cycle, which is why this is the right next measurement rather than a bigger one.

Next: deploy, then wait for dispatch load. The stalls only appear when the scheduler is busy — the external probe saw zero in its first 65 samples at 2 running tasks and then caught 28.58 s and >=30 s once it filled up.

## Deployed, and it answered on the first burst

Deployed 12:55Z. Within four minutes the instrument produced what seven passes of external probing could not.

| blocked | on-CPU | fs read (512B blk) | fs write (512B blk) |
|---|---|---|---|
| 27,961 ms | 116.8% | 8,560 | **2,905,360** |
| 32,642 ms | 120.5% | 0 | **2,905,624** |
| 31,953 ms | 143.0% | 0 | **2,905,360** |
| 28,279 ms | 114.3% | 0 | **2,901,640** |
| 6,486 ms | 125.9% | 206,600 | 16 |
| 5,738 ms | 130.3% | 194,408 | 0 |
| 5,946 ms | 105.9% | **0** | **0** |

**Two distinct classes, not one.**

1. **~28-33 s, ~1.487 GB written.** The write figure repeats to within 0.13% across four occurrences — that is one deterministic operation, not variable load.
2. **~4-6 s, little or no I/O.** One had `fs r0/w0` with 105.9% on-CPU: pure computation.

### What writes 1.487 GB, repeatedly

`/work/cgf-portal` is **11 GB**, of which `.git` is 24 MB and `worktree/` is 10.4 GB. Each worktree is ~435 MB, and the breakdown is one directory: `apps/` at 443 MB → `apps/pipelines` at 440 MB.

That is the per-worktree Python venv. `link-local-assets.sh` deliberately refuses to share it ("**Never add `.venv` here**" — a linked venv makes a worktree import the main checkout's `src/`, and a guard test kept passing against source the worktree never changed). So every worktree builds its own:

- **23 of 27 worktrees carry `apps/pipelines/.venv`**
- **total venv footprint: 9,526 MB**, ~414 MB each

1.487 GB ÷ 414 MB ≈ **3.6 venvs**. A burst of three-to-four concurrent `uv sync` builds accounts for the repeating write signature and its ~30 s duration.

### Correction to the instrument, from its own production data

Two design decisions made against macOS measurements do not hold on the daemon's platform:

- **The fs counters are populated on Linux and are the most informative field in the line.** On macOS `getrusage` reported `fsRead: 0` across 12 `readFileSync` calls, which is why the classifier was narrowed to ignore them. In production they carry the whole finding. They should be promoted from "raw evidence" to a classification input, with the platform difference documented rather than assumed away.
- **`on-CPU %` exceeds 100 (up to 143%), so it is process-wide across threads, not main-thread.** The verdict text says "synchronous JS", which overclaims: CPU burned on libuv threadpool threads while the main thread waits reads identically. The wording needs to say process-wide, and the main thread's own share needs a separate signal.

Neither invalidates the split the instrument was built for — cpu-bound versus parked — which held and produced the finding above.

### Container ceilings raised on the same pass

`cpus: 4.0 → 8.0`, `mem_limit: 16g → 24g`. The compose comment recorded the ceiling as "half the box's 8 cores"; the box has **16**, so the daemon had been running on a quarter. Throttling was real but small — `nr_throttled 29`, `throttled_usec 3.1 s` since boot — so this removes a contributing factor, not the cause. Memory was never the constraint: peak 4.45 GB against a 16 GB cap. The host genuinely is shared (18 containers plus a QEMU VM at 230%, load 5.6-8.4 of 16), so half the box remains the policy.

Stalls continued after the raise (28,279 ms at 2,901,640 blocks), confirming the ceiling was not the mechanism.

## The stall's writer is the daemon process itself, and the working set is tiny

Correcting an attribution I made on AGT-4084 and had started to build on.

`process.resourceUsage()` is `getrusage(RUSAGE_SELF)` — **child processes are excluded**. Everything spawned by the daemon (`uv`, `git`, adapter CLIs) writes outside this counter. So the ~1.49 GB the monitor reports per stall is written by **the Node process itself**.

Confirmed independently from `/proc/7/io`, which needs no interpretation:

```
write_bytes  3,113,492,480 → 3,363,479,552 over 20 s   (+250 MB)
read_bytes         229,376 → unchanged                 (reads nothing)
```

3.36 GB total three minutes after start ≈ two stalls × 1.49 GB. And a 20 s sample taken between stalls showed **+7 MB** — the writes exist only during the blocked windows.

## The constraint this puts on any remaining hypothesis

Every file modified anywhere under `/home/openswarm/.openswarm` and `/work` in a 3-minute span totals **23 MB** (largest: `automation.db` 9 MB, a git pack 6 MB, the WAL 4 MB). Against ~1.5 GB written.

So the daemon rewrites a **few tens of megabytes, hundreds of times over, inside a 28-second synchronous window** — while reading essentially nothing from disk (229 KB since boot) and burning 115-145% of a core.

Ruled out by direct measurement this pass: venv rebuilds (none occurred, and they are subprocesses regardless), knowledge-graph saves (0 in window), `coordination.json` and `task-state.json` atomic rewrites (inode sampled at 10 Hz for 20 s — 0 rewrites of either).

That leaves the SQLite layer as the strongest remaining candidate — `automation.db` is 9 MB with a 4 MB WAL, it is the one thing modified in every window, and better-sqlite3 is fully synchronous on the main thread, which fits both the CPU profile and the blocking. A transaction pattern that rewrites the same pages repeatedly, or checkpoints in a loop, would produce exactly this shape.

Next measurement: sample `/proc/7/io` *inside* a stall rather than around it, and correlate with `automation.db-wal` size and SQLite checkpoint activity. Not asserting SQLite yet — four hypotheses have now died here, and this one has not been tested.

## Localized to the CodeQL security-audit path

Two more candidates died and the writer is now located, though not yet fully explained.

**SQLite is out.** Sampled the write rate at 2 Hz for 150 s alongside the file sizes. Bursts of **194-262 MB/s** recur, and throughout every one of them:

| | value during all bursts |
|---|---|
| `automation.db` | 9,028 KB — unchanged |
| `automation.db-wal` | 4,055 KB — unchanged |
| `coordination.json` | 1,355 KB — unchanged |

200 MB/s written while the database, its WAL and the board never move.

**The earlier file scan was too narrow.** It covered `/home/openswarm/.openswarm` and `/work` only. Scanning the whole container filesystem:

```
1,430 files modified in 2 minutes, all under
/tmp/openswarm-security-audit-6Df3zv/.codeql-javascript/
  trap/javascript/...*.trap.gz        (CodeQL extractor output)
  src/...                             (a full copy of the source tree)
```

`securityAudit.ts:252 createSnapshot` copies every source file into a `mkdtemp` staging root before invoking CodeQL. The live snapshot measured **141 MB**.

**Timing lines up.** Container generation started 13:31:06.

| stall | audit activity |
|---|---|
| 13:31:50, 13:32:43 | first audits after boot |
| 13:35:36 | CodeQL baseline completed 13:34:52 (AX-1030) |
| 13:40:35 | next snapshot root created 13:41:03; baseline 13:37:44 (AX-1044) |

## What is still not explained

`createSnapshot` uses the **async** `fs/promises.cp`, per file, in a sequential loop — that should not block the event loop at all, and 141 MB per snapshot does not obviously add to the ~1.49 GB seen per stall. Two threads of that remain open:

1. **Why the loop blocks.** The profile is >100% on-CPU (115-145%), which means multiple threads — consistent with libuv's threadpool saturated by thousands of queued `cp`/`lstat`/`mkdir` calls. Whether that alone starves a *timer* for 28 s needs testing, not assuming; `UV_THREADPOOL_SIZE` is at its default 4.
2. **Where 1.49 GB comes from** when the snapshot is 141 MB. Concurrent audits across worktrees would multiply it (two were observed), and CodeQL's own extractor writes far more than it copies — but its writes are a subprocess's and should not appear in `/proc/7/io` at all.

## Next measurement

Sample `/proc/7/io` and `/proc/7/status` (`voluntary_ctxt_switches`, thread count) *inside* a single stall while tailing the audit's staging directory, so the write and the staging growth can be compared directly rather than correlated by timestamp. If they match, the fix is to stage by hardlink or reflink instead of copy — the snapshot and the worktree are on the same mount, so `link()` is available and would make the staging nearly free.

Candidates eliminated so far: the `/api/stats` handler, knowledge-graph JSON cost, knowledge scans, embedding inference, the preflight PR-overlap query, the lifecycle lock, the board parse, venv rebuilds, knowledge-graph saves, state-file atomic rewrites, SQLite.

## The writer is the main thread, synchronously — and it is none of the files

Four more candidates died this pass, including one I had wrongly declared dead and one I had wrongly declared alive.

**The security audit is not it.** Sampled the daemon's write rate against the audit staging directory's growth, once a second. They are *anti-correlated*:

| seconds | daemon | staging |
|---|---|---|
| 122-133 | **45-54 MB/s** | **0 MB/s** |
| 142-146 | ~0 MB/s | 8-11 MB/s |

CodeQL's extraction is a subprocess and shows up exactly where it should — in the staging directory, not in `/proc/7/io`. It explains the 1,430 files in `/tmp`; it does not explain the daemon's writes.

**SQLite is properly out now.** I had ruled it out on file *size*, which was the wrong test — a WAL in reuse mode is rewritten in place and never grows. Re-tested on write volume: WAL mtime changed **3 times in 20 s** while `write_bytes` grew **616 MB**. Three touches of a 4 MB WAL is at most 12 MB of 616.

Settings, for the record: `journal_mode=wal`, `synchronous=1 (NORMAL)`, `wal_autocheckpoint=1000`, `page_size=4096`.

**Not the embedding stack.** Loaded exactly **once**, at 13:32:48. The `onnxruntime`/`lancedb` mtimes I had flagged as "recently modified" are the image build time (20:06 KST); that reading came from a malformed `du` aggregation in my own probe, not from the filesystem.

**Not any file, by elimination.** The daemon's open file descriptors are, in full: `automation.db`, its `-wal` and `-shm`, `registry.db` and its pair, `service-instance-lock.db`, and `/dev/null`. Nothing large is open for writing. A 2-second-resolution mtime scan of the whole state directory shows only the expected occasional touches (WAL, `coordination.json`, `task-state.json`, `ci-state.json`) — never at 50 MB/s.

## The new constraint, which is the useful part

Per-thread accounting (`/proc/7/task/*/io`): the writes are attributed to **tid 7 — the main thread**, not the libuv threadpool.

So this is **synchronous fs from JavaScript**, in bursts, on the thread that runs the event loop. That is both consistent with the 28-33 s blocking and a much sharper target than "something writes 1.5 GB": it is a `writeFileSync`/`appendFileSync`-class call, executed enough times in one pass to move gigabytes, against a working set of tens of megabytes.

## Next step, and why it has to be in-process

`strace` is unavailable and the container has no `CAP_SYS_PTRACE` (`CapAdd=[]`, `Privileged=false`), so syscall tracing would need the container's capabilities changed — not something to do unilaterally on a live daemon.

The equivalent from inside: extend the event-loop monitor to wrap `fs.writeFileSync`, `fs.appendFileSync` and `fs.writeSync`, accumulating bytes and a captured call-site frame, and have a stall report name its top three writers. Same shape as the monitor itself — an instrument that answers the question the log cannot — and it would close this outright rather than eliminating a fifteenth candidate.

Eliminated so far: `/api/stats` handler · knowledge-graph JSON cost · knowledge scans · embedding inference · embedding reloads · preflight PR-overlap query · lifecycle lock · board parse · venv rebuilds · knowledge-graph saves · state-file atomic rewrites · SQLite (twice, properly the second time) · CodeQL staging · overlayfs copy-up of native modules.

## Built the write-attribution instrument, then measured that its delivery mechanism cannot work

Wrote `syncWriteTracker` — wraps `fs.writeFileSync`/`appendFileSync`/`writeSync`, tallies bytes per destination, and appends the top writers to a stall line. 30 tests, 5 mutations killed. Two review findings found and fixed on the way:

- **`writeSync` slice accounting.** `writeSync(fd, buffer, offset, length)` writes a slice, and pooled buffers are much larger than the slice. Measuring the whole buffer overstates the volume — which is this tool's only output.
- **Overlapping lifetimes.** A second `start` captured the first's wrappers as its "originals"; the later `stop` would then restore a wrapper and leave `fs` permanently patched. Now single-owner.

**Then the third finding killed the delivery.** Named ESM imports bind at link time and never see a patch applied afterwards. I had noted that as a caveat in the code; the reviewer was right that it is not a caveat but the deciding fact, so I counted it:

```
src/ files calling a synchronous fs write (non-test): 15
  named ESM import  →  11    ← invisible to the patch
  other             →   4
```

`worktreeManager.ts` (4 calls), `cli.ts`, `cli/daemon.ts`, `telemetry.ts`, `memoryMcp.ts`, `modelCatalog.ts`, `embeddingConfig.ts`, `updateNotifier.ts`, `designPipeline.ts`, `support/dev.ts`, `tui/inputDebug.ts` all import by name. An instrument blind to 11 of 15 call sites would have produced an empty tally and looked like *evidence of absence*.

Parked on `swarm/AGT-4079-sync-write-attribution` (`af18663`), not shipped.

## The shape that would work

A `--require` preload that patches CJS `fs` **before any ESM module links**, which is how OpenTelemetry-style instrumentation reaches named imports. That changes how the daemon is launched (Dockerfile `CMD`), so it is a deliberate step rather than a drop-in — but the pure parts of this branch (`payloadBytes`, `writeSyncBytes`, `recordWrite`, `topWriters`) are correct and tested regardless of delivery, so the rework is the patching layer only.

## Standing state

The stall is unchanged and reproducible — roughly one 28-34 s block per minute under dispatch load, each writing ~1.49 GB from the **main thread**, against a working set of tens of megabytes, reading essentially nothing.

Fifteen mechanisms eliminated by measurement so far: `/api/stats` handler · knowledge-graph JSON cost · knowledge scans · embedding inference · embedding reloads · preflight PR-overlap query · lifecycle lock · board parse · venv rebuilds · knowledge-graph saves · state-file atomic rewrites · SQLite (twice) · CodeQL staging · overlayfs copy-up of native modules.

### 2026-08-30T03:02:10.045Z
**Measurement at the exact condition this issue names — concurrency 9 — taken 2026-08-30T03:01Z.**

The loop reached 9/9 concurrent EXECUTING runs for the first time after three vega repos were enabled. Container state at that moment:

```
health=healthy  restarts=0
cpu=301.62%  mem=3.567GiB / 24GiB (14.86%)

/api/health latency:  0.51 s, 0.43 s, 0.01 s   (HEALTHCHECK timeout is 10 s)
last 5 healthcheck exits: 0 0 0 0 0
```

So the flap is **not reproducing** here: probes are two orders of magnitude under the timeout, CPU is at 301% of an 800% ceiling, and memory is at 15% of its limit — nowhere near the 86% that AGT-4062 recorded.

**Do not read this as resolved.** Two reasons to hold it open:

1. The nine runs had only just started — they were in draft/worker stages, not yet at the CodeQL and verify stages that AGT-4062 identified as the starving load. The heavy part of the window had not arrived.
2. Event-loop stalls *are* still firing in the same period, from the always-on monitor: 2.2 s, 4.7 s, 4.8 s, 6.5 s, 6.6 s, 7.3 s — all `cpu-bound`, synchronous JS, one at **141.3% of the window** (12482 ms user CPU in a 9048 ms window, i.e. consuming more than one core). The fs-read counts split cleanly into two groups — 68k–101k reads on some, `fs r0/w0` on others — so at least two different synchronous paths are involved, not one.

A 7 s stall is well short of the 10 s HEALTHCHECK timeout, which is consistent with both observations: the stalls are real and ongoing, but currently below the threshold that flaps the container. Whether they cross it under the verify/CodeQL phase at this concurrency is the measurement still missing, and the loop is now in a state where it can be taken.

### 2026-08-30T03:03:27.292Z
**Follow-up 45 minutes later — the stalls do not track worker concurrency. They track restart.**

My previous comment said the heavy verify/CodeQL phase had not yet arrived and the measurement was still missing. It has now arrived, and it reverses the expected direction.

| Window | What was running | `[EventLoop] Blocked` events |
| -- | -- | -- |
| 02:52:31–02:56 | container restart, first heartbeats (793 tasks fetched, task-state loaded) | **6** — 2.2 s, 4.7 s, 4.8 s, 6.5 s, 6.6 s, 7.3 s |
| 02:56–03:04 | 9/9 concurrent runs, 3 CodeQL baselines completed, verify + security-audit started | **0** |

Container at the second window: `healthy`, RestartCount 0, CPU 203.96% of an 800% ceiling, memory 19.93%. Not one stall while the load this issue predicts would cause them was actually present.

So the framing "spikes past the timeout **at concurrency 9**" does not hold as a causal story. Every stall observed in this session sat in the minutes after a container restart, while the daemon was loading its own state — consistent with the two fs-read groups noted earlier (68k–101k reads on some, `fs r0/w0` on others: file enumeration and a large synchronous parse, not worker activity). `task-state.json` is 1.9 MB and 793–799 Linear tasks are enriched per heartbeat.

Worth noting alongside it: heartbeats are now firing every 15–30 s (02:55:47, 02:56:15, 03:00:00, 03:01:06, 03:01:54, 03:02:19, 03:02:34, 03:02:47) rather than on the `*/5 * * * *` cron, because `scheduleNextHeartbeat()` re-triggers after each completion. Each one re-fetches the full task list. That did **not** produce stalls here — the Linear client's `myIssuesCache` absorbs the repeats — but it means the startup window is the one time that work is uncached, which fits the correlation exactly.

**What this changes:** if the goal is to stop the container flapping unhealthy, the window to instrument is the first ~4 minutes after a restart, not sustained high concurrency. I have not seen the flap itself in this session (health probes measured 0.51 s / 0.43 s / 0.01 s against a 10 s timeout), so I cannot say whether the startup stalls ever cross it — only that they are the ones that get close, at 7.3 s.

- **PR #581 (swarm/AGT-4008-feature-implement-authenticated-github-r)** — 2 file(s): `package-lock.json`, `src/support/web.ts`
- **PR #580 (swarm/AGT-3417-fix-async-ui-contain-rejected-commands-a)** — 1 file(s): `package-lock.json`
- **PR #579 (swarm/AGT-3492-fix-network-access-restrict-notification)** — 2 file(s): `package-lock.json`, `src/support/web.ts`

## Linear
Closes AGT-4079

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -452,6 +452,40 @@
         "url": "https://github.com/discordjs/discord.js?sponsor"
       }
     },
+    "node_modules/@emnapi/core": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
+      "integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "@emnapi/wasi-threads": "1.2.1",
+        "tslib": "^2.4.0"
+      }
+    },
+    "node_modules/@emnapi/runtime": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
+      "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "tslib": "^2.4.0"
+      }
+    },
+    "node_modules/@emnapi/wasi-threads": {
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.1.tgz",
+      "integrity": "sha512-uTII7OYF+/Mes/MrcIOYp5yOtSMLBWSIoLPpcgwipoiKbli6k322tcoFsxoIIxPDqW01SQGAgko4EzZi2BNv2w==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "tslib": "^2.4.0"
+      }
+    },
     "node_modules/@envelop/core": {
       "version": "5.5.1",
       "resolved": "https://registry.npmjs.org/@envelop/core/-/core-5.5.1.tgz",
@@ -2022,9 +2056,9 @@
       }
     },
     "node_modules/@jridgewell/sourcemap-codec": {
-      "version": "1.6.0",
-      "resolved": "https://registry.npmjs.org/@jridgewell/sourcemap-codec/-/sourcemap-codec-1.6.0.tgz",
-      "integrity": "sha512-T7jf+5zgsZHwNJ4lvQ7/aezbyk0nNX+zJVWpmHA7VYsEx7a7qr5Rg5IbtJFqkgze5Y2sruq1RUY8Q837Od7iFw==",
+      "version": "1.5.5",
+      "resolved": "https://registry.npmjs.org/@jridgewell/sourcemap-codec/-/sourcemap-codec-1.5.5.tgz",
+      "integrity": "sha512-cYQ9310grqxueWbl+WuIUIaiUaDcj7WOq5fVhEljNVgRfOUhY9fy2zTvfoqWsnebh8Sl70VScFbICvJnLKB0Og==",
       "dev": true,
       "license": "MIT"
     },
@@ -2249,21 +2283,33 @@
         }
       }
     },
-    "node_modules/@napi-rs/lzma-linux-x64-gnu": {
-      "version": "1.5.1",
-      "resolved": "https://registry.npmjs.org/@napi-rs/lzma-linux-x64-gnu/-/lzma-linux-x64-gnu-1.5.1.tgz",
-      "integrity": "sha512-oTXEIha4SsuXdTA4Iyskj0kpdx2yVXdhd75c2v3xGrHFfVMsbhTPZU/nMPL4sWKo4pBHm3aucLaqGlF696dTyQ==",
-      "cpu": [
-        "x64"
-      ],
+    "node_modules/@napi-rs/wasm-runtime": {
+      "version": "1.1.4",
+      "resolved": "https://registry.npmjs.org/@napi-rs/wasm-runtime/-/wasm-runtime-1.1.4.tgz",
+      "integrity": "sha512-3NQNNgA1YSlJb/kMH1ildASP9HW7/7kYnRI2szWJaofaS1hWmbGI4H+d3+22aGzXXN9IJ+n+GiFVcGipJP18ow==",
       "dev": true,
       "license": "MIT",
       "optional": true,
-      "os": [
-        "linux"
-      ],
-      "engines": {
-        "node": "^22.20 || ^24.12 || >=25"
+      "dependencies": {
+        "@tybys/wasm-util": "^0.10.1"
+      },
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/Brooooooklyn"
+      },
+      "peerDependencies": {
+        "@emnapi/core": "^1.7.1",
+        "@emnapi/runtime": "^1.7.1"
+      }
+    },
+    "node_modules/@oxc-project/types": {
+      "version": "0.133.0",
+      "resolved": "https://registry.npmjs.org/@oxc-project/types/-/types-0.133.0.tgz",
+      "integrity": "sha512-KzkdCd6Uxqnf6l3HOw1xfatAlUURA0g14cvBYFyJ5SaNOQbOUvBr9PKArcPcrNIeRsBdgcUzOGrhKveVpvOIGA==",
+      "dev": true,
+      "license": "MIT",
+      "funding": {
+        "url": "https://github.com/sponsors/Boshen"
       }
     },
     "node_modules/@oxlint/darwin-arm64": {
@@ -2441,24 +2487,10 @@
       "integrity": "sha512-Javneu5lsuhwNCryN+pXH93VPQ8g0dBX7wItHFgYiwQmzE1sVdg5tWHiOgHywzL2W21XQopa7IwIEnNbmeUJYA==",
       "license": "MIT"
     },
-    "node_modules/@rollup/rollup-android-arm-eabi": {
-      "version": "4.63.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.63.1.tgz",
-      "integrity": "sha512-UZ8sUxPTiHWYX9QNdJedb1kDZSpS1t/VPWBWGSgqHNi9w3Cu6IXvu2mzbhiTiPvtrqgTQJ+zqiAq2iPIPilpaQ==",
-      "cpu": [
-        "arm"
-      ],
-      "dev": true,
-      "license": "MIT",
-      "optional": true,
-      "os": [
-        "android"
-      ]
-    },
-    "node_modules/@rollup/rollup-android-arm64": {
-      "version": "4.63.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.63.1.tgz",
-      "integrity": "sha512-cQ4nFQABN5cDvDpbvJ7bMStCpnaVxynZrRMfUJYgxcIk9Sh54FIO1vtfkg0B69REjER77ioZ/ov+eAApx/KmLQ==",
+    "node_modules/@rolldown/binding-android-arm64": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-android-arm64/-/binding-android-arm64-1.0.3.tgz",
+      "integrity": "sha512-454rs7jHngixp/NMxd5srYD57OnzSlZ/eFTETjORQHLwJG1lRtmNOJcBerZlfu4GjKqeq8aCCIQrMdHyhI51Hw==",
       "cpu": [
         "arm64"
       ],
@@ -2467,12 +2499,15 @@
       "optional": true,
       "os": [
         "android"
-      ]
+      ],
+      "engines": {
+        "node": "^20.19.0 || >=22.12.0"
+      }
     },
-    "node_modules/@rollup/rollup-darwin-arm64": {
-      "version": "4.63.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-darwin-arm64/-/rollup-darwin-arm64-4.63.1.tgz",
-      "integrity": "sha512-FQNqd1lRy/0QhDk3xeRIkSBiCpXCiDnZO3YLVdcDKN1UBiKToNftCzcXYNLshmPDUMlu2TdeS8tGcsU6f3YF1Q==",
+    "node_modules/@rolldown/binding-darwin-arm64": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-darwin-arm64/-/binding-darwin-arm64-1.0.3.tgz",
+      "integrity": "sha512-PcAhP+ynjURNyy8SKGl5DQP94aGuB/7JrXJb/t7P+hanXvQVMWzUvRRhBAcg/lNRadBhoUPqSoP4xw5tR/KBEA==",
       "cpu": [
         "arm64"
       ],
@@ -2481,12 +2516,15 @@
       "optional": true,
       "os": [
         "darwin"
-      ]
+      ],
+      "engines": {
+        "node": "^20.19.0 || >=22.12.0"
+      }
     },
-    "node_modules/@rollup/rollup-darwin-x64": {
-      "version": "4.63.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-darwin-x64/-/rollup-darwin-x64-4.63.1.tgz",
-      "integrity": "sha512-pvD16V939D3CloK0+qikpGaxiPrDUXTe7Y5cWOMkMSy7m1cawa8EGy/kXYi/G/cKAC4HDAbSnzCIk1WmsoOKXg==",
+    "node_modules/@rolldown/binding-darwin-x64": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-darwin-x64/-/binding-darwin-x64-1.0.3.tgz",
+      "integrity": "sha512-9YpfeUvSE2RS7wysJ81uOZkXJz7f7Q55H2Gvp3VEw/EsahqDtrphrZ0EwDLK5vvKOzaCrBsjF8JmnMLcUt78Gg==",
       "cpu": [
         "x64"
       ],
@@ -2495,26 +2533,15 @@
       "optional": true,
       "os": [
         "darwin"
-      ]
-    },
-    "node_modules/@rollup/rollup-freebsd-arm64": {
-      "version": "4.63.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-freebsd-arm64/-/rollup-freebsd-arm64-4.63.1.tgz",
-      "integrity": "sha512-pcFGeL2345VwdTnJhA6zLbew+YgWB0qBG2+dMtXjCicf6+rm6kO6cOoh5VnTe0ZMrMRgRyuHmCJxZWrIdzYuOw==",
-      "cpu": [
-        "arm64"
       ],
-      "dev": true,
-      "license": "MIT",
-      "optional": true,
-      "os": [
-        "freebsd"
-      ]
+      "engines": {
+        "node": "^20.19.0 || >=22.12.0"
+      }
     },
-    "node_modules/@rollup/rollup-freebsd-x64": {
-      "version": "4.63.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-freebsd-x64/-/rollup-freebsd-x64-4.63.1.tgz",
-      "integrity": "sha512-mRJlqSRulVzcKq/LKA6ICSIc3K/l4fzlVn/gePn2nXIHy8seRi5z/eeRE0d/XMBxcMldiXtQTSpRj0tkkC3g8Q==",
+    "node_modules/@rolldown/binding-freebsd-x64": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-freebsd-x64/-/binding-freebsd-x64-1.0.3.tgz",
+      "integrity": "sha512-yB1IlAsSNHncV6SCTL27/MVGR5htvQsoGxIv5KMGXALp+Ll1wYsn+x98M9MW7qa+NdSbvrrY7ANI4wLJ0n1e6g==",
       "cpu": [
         "x64"
       ],
@@ -2523,12 +2550,15 @@
       "optional": true,
       "os": [
         "freebsd"
-      ]
+      ],
+      "engines": {
+        "node": "^20.19.0 || >=22.12.0"
+      }
     },
-    "node_modules/@rollup/rollup-linux-arm-gnueabihf": {
-      "version": "4.63.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-linux-arm-gnueabihf/-/rollup-linux-arm-gnueabihf-4.63.1.tgz",
-      "integrity": "sha512-YDUNvVM85TI3g/1OpnqKP1h4NeW/j64DfWMf+G3M809xNk1bJSnpFp4sh83NpmVE5DXnkh8ULor4LTVZKoYLHw==",
+    "node_modules/@rolldown/binding-linux-arm-gnueabihf": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-linux-arm-gnueabihf/-/binding-linux-arm-gnueabihf-1.0.3.tgz",
+      "integrity": "sha512-Yi30IVAAfLUCy2MseFjbB1jAMDl1VMCAas5StnYp8da9+CKvMd2H2cbEjWcw5NPaPqzvYkVIaF1nNUG+b7u/sw==",
       "cpu": [
         "arm"
       ],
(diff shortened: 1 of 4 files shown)


unohee:
## 🔍 Fresh review of 4ef4836 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  건강진단 엔드포인트 구조/테스트/배선 모두 깔끔. 유일한 지적은 getVersion()이 매 초 package.json을 동기 읽는 점 — 모듈 로드 시 한 번만 읽고 캐싱하는 게 좋음.

PR #585, opened 2026-09-09
[Bug] coordinationTools test asserts an uncoloured substring — green in CI, red on every developer machine

unohee:
## Problem

`src/coordination/coordinationTools.test.ts` → *"a wait keeps its process alive until it settles > resolves in a child process with nothing else pending"* fails locally:

```
AssertionError: expected 'SETTLED [] <ESC>[33mtrue<ESC>[39m\n' to contain 'SETTLED [] true'
```

The child process writes `true` wrapped in a yellow ANSI colour code. The assertion looks
for the bare substring, so it only passes where stdout is **not** a TTY — which is exactly
CI. The result is a test that is permanently green in CI and permanently red for anyone
running the suite on their own machine.

## Evidence it is pre-existing, not from any current branch

Reproduced on pristine `origin/main` at `9704ffe` by stashing all local changes in a fresh
worktree and running only that file: **1 failed | 22 passed**. It reproduces the same way
with an unrelated locale-only patch applied, so nothing in flight causes it.

## Why it matters more than one red test

It trains people to ignore a failing suite. This session lost time twice deciding whether
a failure belonged to the change in hand — which is precisely the cost a test suite exists
to avoid. It also means the assertion is not actually testing what it reads as: it passes
in CI for the wrong reason.

## Scope

* Make the assertion independent of colour — strip ANSI from the captured output before
  matching, or spawn the child with colour disabled (`FORCE_COLOR=0` / `NO_COLOR=1`) so the
  captured text is stable in both environments.
* Prefer whichever keeps the test asserting the *behaviour* (the wait settled) rather than
  the exact rendering.
* Check for sibling tests in this file that capture child-process stdout and would fail the
  same way once someone runs them on a TTY.

## Done

* The test passes both with and without a TTY.
* The assertion no longer depends on whether colour is enabled.
* A short comment records why, so it is not "simplified" back.

- **PR #584 (swarm/AGT-4059-bug-dashboard-s-latestaddressable-can-ta)** — 1 file(s): `package-lock.json`
- **PR #583 (swarm/AGT-4079-bug-api-health-spikes-past-the-10s-healt)** — 1 file(s): `package-lock.json`
- **PR #581 (swarm/AGT-4008-feature-implement-authenticated-github-r)** — 1 file(s): `package-lock.json`
- **PR #580 (swarm/AGT-3417-fix-async-ui-contain-rejected-commands-a)** — 1 file(s): `package-lock.json`
- **PR #579 (swarm/AGT-3492-fix-network-access-restrict-notification)** — 1 file(s): `package-lock.json`

## Linear
Closes AGT-4153

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1300,9 +1300,6 @@
       "cpu": [
         "arm"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -1318,9 +1315,6 @@
       "integrity": "sha512-0DaL0A6Xu6sQSQFwe4iVCrKWU2cCTItnRsYsCdxAMm9NF6twAA9BKnoqy4hqz4+azQ0JHuA26qiUKsf1XJ/v5A==",
       "cpu": [
         "arm64"
-      ],
-      "libc": [
-        "glibc"
       ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
@@ -1338,9 +1332,6 @@
       "cpu": [
         "ppc64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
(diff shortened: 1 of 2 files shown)


unohee:
## 🔍 Fresh review of 9f9f621 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  package-lock.json은 무해한 lockfile regeneration, 테스트 수정은 chalk ANSI 컬러로 인한 CI/로컬 불일치를 정확히 해결. blocking issue 없음.
  Suggestions:
    - NO_COLOR만 있어도 충분하나 FORCE_COLOR도 함께 설정된 것은 방어적으로 안전하므로 유지해도 무방

PR #694, opened 2026-09-19
feat(runtime): make test parallelism resource-aware

unohee:
## Summary
- compute a conservative per-command test worker budget from CPU load and free memory
- propagate caps to delegated CLIs, native bash tools, strict sandbox execution, and deterministic verification
- clamp explicit Vitest/Jest, pytest, and Cargo worker overrides while preserving stricter operator limits

## Verification
- 223 focused tests passed under `OPENSWARM_TEST_PARALLELISM=1`
- `npm run build`
- `npm run lint` (one pre-existing warning in `web/static/js/usage.mjs`)
- OpenSwarm review iterated on all substantiated findings; final ordinary-bash finding was directly disproved by `env: buildBashToolEnv()` plus focused tests

Linear: AGT-4475


unohee:
## 🔍 Fresh review of c9ac59b (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  All four execution paths (sandbox executor in agenticLoop, sandbox executor in verify runner, direct execFileAsync in bash tool, worker env via envPath) correctly call into testResourceBudget. The command-rewrite logic handles shell sequences, cd, comments, env overrides, and the various test runners (vitest, jest, pytest, cargo). The only risk is double-budgeting in the verify runner where `resourceAwareTestCommand` and `safeEnv` construction both clamp the same values independently, but they agree on the budget so the outcome is consistent. No material defect found.
  Suggestions:
    - Consider testing subshell-wrapped commands like `$(npm test)` — the splitter won't recurse into them, but that's an acceptable conservative gap.
    - In runner.ts line 391, `resourceAwareTestCommand(command.run, cwd, undefined, env)` uses the raw caller env; line 396 onward builds a safeEnv from it with the same snapshot. This works because both call `effectiveTestParallelism` on the same `OPENSWARM_TEST_PARALLELISM` value, but a reader might wonder why the same computation runs twice on slightly different env shapes.

PR #695, opened 2026-09-19
feat(memory): stage agent lessons until a run succeeds (AGT-4461)

unohee:
Summary: adds staged remember entries, promotes only on successful completion with provenance, discards rollback iteration entries, and adds memory preview. Verification: focused Vitest suites, typecheck, lint, build. Closes AGT-4461.


unohee:
## 🔍 Fresh review of 07538c8 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  All wiring is correct — `memoryContext` flows through every adapter, `remember` is properly filtered in readOnly/no-scratchpad modes, and promotion only fires on green completion. The atomic write pattern (tmp+rename) avoids partial writes. One design note worth internalising: `remember.json` accumulates across attempts because the scratchpad key is stable, which matches the existing `scratch_write` convention.

PR #700, opened 2026-09-19
fix(service): survive revoked Linear OAuth without identity fallback (AGT-4455)

unohee:
## Summary

- Keep the service running when the preferred Linear OAuth profile is revoked or cannot refresh.
- Leave Linear disabled for that service run rather than silently falling back to the API key, which can represent a different actor.
- Log the degraded Linear state and the deliberate identity-preserving choice.

## Verification

- `npx vitest run src/core/service.test.ts --maxWorkers=1` (51 passed)
- `npm run typecheck`


unohee:
## 🔍 Fresh review of bdb8a10 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Tested the catch block end-to-end: mocked AuthProfileStore returns a profile, ensureValidToken rejects, the daemon stays up without Linear, warn log contains both key strings. The control flow correctly prevents fall-through to the API-key branch because it's an else-if that the outer if already matched. No defects found.

PR #703, opened 2026-09-19
fix: record CLI adapter sessions

unohee:
## Summary

- Record delegated CLI adapter invocations at the observable process boundary: start, prompt, raw stdout/stderr, and terminal outcome.
- Mark these records as `recordingLevel: "cli"` so they cannot be mistaken for the richer native-loop turn transcript.
- Carry adapter/task/stage attribution into the recorder, with process-context compatibility for existing callers.

## Verification

- `npm test -- --run src/adapters/base.spawn.test.ts src/adapters/agenticLoop.sessionLog.test.ts`
- `npm run typecheck`
- `npm run lint` (one existing warning in `web/static/js/usage.mjs`)

Daemon was intentionally kept stopped; no runtime service was started.


unohee:
## 🔍 Fresh review of 6ede503 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  CLI-level session recording is correctly wired: `recordingLevel: 'cli'` is typed, scoped to the process boundary, and the fallback chain (`usageAttribution → processContext → adapter.name`) handles all 11 callers. The idempotent close in sessionLog.ts makes the finally-block double-close safe. The env-var restore pattern in tests is thorough (hoisted capture + beforeEach/afterEach). The only gap is that the finally block always passes `outcome: 'threw'` even on success — no data corruption, but misleading if anyone inspects the close payload. Consider guarding it with a caught/uncaught flag.
  Suggestions:
    - Guard the finally-block close with a `let caught = true; ... caught = false; finally { if (caught) session?.close(...) }` pattern so the outcome label reflects what actually happened, not a default placeholder.

PR #706, opened 2026-09-19
fix: record watchdog timeout in ledger

unohee:
## Summary
- persist scheduler hard-watchdog aborts as `watchdog_timeout` in the durable attempt row
- retain the derived budget and actual elapsed wall-clock in the diagnostic message
- leave unrelated `infra_error` attempts unchanged

## Verification
- `npm test -- --run src/automation/autonomousRunner.infraError.test.ts src/automation/durableRunCoordinator.failureDetail.test.ts` (18 passed)
- `npm run typecheck`


unohee:
## 🔍 Fresh review of 3d28612 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  The diff wires the scheduler's hard watchdog timeout into the durable ledger via `recordWatchdogTimeout`. The `activeClaims` map allows the error handler to find the correct claim before the executor terminates. SQL-level guards (`result_status IS NULL` + `state_version` fencing) prevent double-recording when both watchdog and executor-catch paths fire. The test correctly use the real `DurableRunCoordinator` with a SQLite database and simulates the production call chain. Verified: regex matches scheduler format (taskScheduler.ts:611), timing arithmetic is correct, activeClaims cleanup in finally block is safe, and claim fence at line 720 prevents stale deletes.

PR #707, opened 2026-09-19
fix: pin durable write scope across retries

unohee:
## Summary
- pin the first durable write scope for a run and preserve it across later task observations
- reconstruct resumed tasks from the pinned scope so publication validates every branch commit against its original contract

## Verification
- `npm test -- --run src/automation/durableRunCoordinator.test.ts` (61 passed)
- `npm run typecheck`


unohee:
## 🔍 Fresh review of a0f4032 (`openswarm pr review --fresh`)

✎ Decision: REVISE
  `metadataConflictScope` in `runLedgerScope.ts:22-25` reads only `fileScope`, but after pinning the stored `fileScope` can be narrower than `pinnedFileScope`. This allows other runs to claim conflicting scopes against a live pinned run. Fix: read `pinnedFileScope ?? fileScope`. Add a cross-run conflict test in the same describe block.
  Suggestions:
    - Add a test that registers run A via observeTask with broad scope, rediscover-narrows it, claims it, then tries to claim run B with a scope that overlaps only the pinned broad scope — it should be rejected.
    - The cosmetic formatting changes (joining `const ledger = ...; const coordinator = ...` onto one line) are unnecessary churn but harmless.

unohee:
## 🔍 Fresh review of 3e7a76b (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Traced the full pin lifecycle — observeTask → registerRun merge → execute admission → metadataConflictScope → runRecordToTask. The pin correctly survives rediscovery narrowing, pre-existing rows fall back to fileScope, and the test exercises the real DB path including conflict rejection. No regressions or bugs found.

PR #710, opened 2026-09-20
feat(action): accept an explicit review model

unohee:
Part of AGT-3099.

Adds the review CLI model override and forwards an optional model Action input through an argv array. When a model is requested, the Action capability-checks the installed CLI, so an older package fails as gate-not-run rather than silently ignoring the cost cap.

The dogfood workflow explicitly pins deepseek/deepseek-v4-flash, enabling the required external-repository Action E2E without invoking an expensive default model.


unohee:
## 🔍 Fresh review of c9ee337 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Wiring is complete and type-safe end-to-end — action.yml → env → CLI option → ReviewCommandOptions → runReviewer → spawnCli → adapter. The version-compatibility guard in action.yml correctly exits 2 if an older CLI lacks `--model`. No broken links or silent fallbacks.
  Suggestions:
    - Add a test in reviewCommand.test.ts that exercises `model` propagation through `runReviewCommand` (e.g., verify it reaches `deps.review` or the default `runReviewer` call). Currently the parameter is untested at the command level.

PR #713, opened 2026-09-20
fix: fail closed on unparsed review output

unohee:
## Summary
- normalize a prefixed OpenSwarm JSON review document before reading action outputs
- fail `fail-on-gate-not-run` whenever no parsed `gateRan: true` verdict exists, including exit code 0

## Evidence
- External cgf-portal E2E run 35479878890 exposed a real false-green: a late memory diagnostic prefixed the JSON document, the action emitted `no changes`, and the job passed despite `gate-ran=false`.
- Reproduced the exact prefixed JSON shape locally; the new extractor recovered `approve` and `gateRan=true`.
- `git diff --check`.


unohee:
## 🔍 Fresh review of 2911fe7 (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Verified action.yml JSON parsing robustness (diagnostic noise, gate-not-run detection, clean-file extraction), intentional LanceDB downgrade for cross-platform npm ci, and CHANGELOG consistency. No defects found in the current diff.
  Suggestions:
    - Consider adding a similar parse-validation guard for the SARIF output path if the CLI can also embed diagnostics in SARIF stdout.

PR #714, opened 2026-09-20
fix: preserve reviewer verdict on salvage

unohee:
## Root cause
A real reviewer session for AX-1366 exhausted 30 tool turns. The generic no-tools salvage prompt only asked for the original task format, so the model returned a worker-style status summary with no `Decision`/JSON. The strict reviewer parser correctly rejected it, producing the recurring AGT-4484 bucket.

## Fix
When salvage ends a review/verdict task, explicitly require `Decision: approve|revise|reject`, concrete reasoning, and forbid a worker-status summary. This applies through the shared agentic loop used by all in-process adapters.

## Verification
- `npx vitest run src/adapters/agenticLoop.test.ts src/adapters/resultParsing.test.ts src/agents/reviewer.test.ts --reporter=dot` — 139 passed
- `npm run typecheck`
- `npm run lint`
- `git diff --check`

Regression test drives tool exhaustion and asserts the salvage prompt contains the explicit verdict contract.


unohee:
## 🔍 Fresh review of a3ddccc (`openswarm pr review --fresh`)

✓ Decision: APPROVE
  Prompt change is correctly conditional (only activates for review tasks), no regression risk. Test validates the prompt text and result propagation. The test relies on source-string inspection (criterion #8) — it would pass if the same logic moved to a helper — but no better behavioral alternative exists at the unit level for this guardrail.

